Issue #13, opened 2018-09-19
Create `IssueWanted.Db.Repo` module with schemas for repository


chshersh:
We want to store some information in our database regarding all Haskell repositories. We don't need to store everything about repo. Only interesting part. Schema of repo table should be created using `squeal` library.

chshersh:
This issue is very vague and broad. To be clarified better in separate issues. Thanks @willbasky for your effort so far!

PR #18, opened 2018-09-21
[#13] Add repo schema

willbasky:
Resolves [#13](https://github.com/kowainik/issue-wanted/issues/13)

### DIFF
--- a/src/IssueWanted/Db/Repo.hs
+++ b/src/IssueWanted/Db/Repo.hs
@@ -4,31 +4,34 @@
 
 module IssueWanted.Db.Repo where
 
-import Squeal.PostgreSQL ((:::), NullityType (..), PGType (..))
-import Squeal.PostgreSQL.Schema (ColumnType (..))
-
-import qualified GHC.Generics as GHC
-
+import Squeal.PostgreSQL ((:::), NullityType (..), PGType (..), (:=>))
+import Squeal.PostgreSQL.Schema ( ColumnConstraint(..), SchemumType(..), TableConstraint(..))
 
 type Schema =
-    '[ "repos"  ::: -- all Haskell repos
-        '[ "repoId"       ::: 'Required ('NotNull 'PGint4)
-         , "repoName"     ::: 'Required ('NotNull 'PGtext)
-         , "repoUrl"      ::: 'Required ('NotNull 'PGtext)
-         , "repoOwner"    ::: 'Required ('NotNull 'PGtext)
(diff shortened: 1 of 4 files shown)


willbasky:
I exported `import Squeal.PostgreSQL.Schema (ColumnType (..))`
but `Not in scope: data constructor ‘Required’`.
Any assumptions?

src/IssueWanted/Db/Repo.hs
```diff
@@ -0,0 +1,34 @@
...
+        '[ "gfiTitle"  ::: 'Optional ('NotNull ' PGtext)
+         , "gfiUrl"    ::: 'Optional ('NotNull 'PGtext)
+         , "gfiRepoId" ::: 'Required ('NotNull 'PGint4)
+        ]
+
+    , "repoIssueHW" ::: -- all Haskell issues labeled "Help Wanted"
+        '[ "hwTitle"  ::: 'Optional ('NotNull 'PGtext)
+         , "hwUrl"    ::: 'Optional ('NotNull 'PGtext)
+         , "hwRepoId" ::: 'Required ('NotNull 'PGint4)
+        ]
```
ChShersh: Looks like I need to read about `squeal` as well :joy: Looks legit at first glance. But need to double check. Also, I saw that `squeal` supports pretty-printing output of SQL query for creating table. Maybe it's a good idea to add it as a comment to this schema type? :thinking:

ChShersh:
@willbasky 

> but Not in scope: data constructor ‘Required’

Probably it's from old version of library. I don't see `Required` in latest version `0.4.0`.

src/IssueWanted/Db/Repo.hs
willbasky: > I saw that squeal supports pretty-printing output of SQL query for creating table. 

Could you share the link?

ChShersh: This section of tutorial contains it:

* https://github.com/morphismtech/squeal#usage

willbasky: Sorry, I couldn`t find pretty section :(

ChShersh: Well, it might not be pretty, just Haskell translated to SQL. Goes after:

> We can easily see the generated SQL is unsurprising looking.

chshersh:
@willbasky thanks for your huge work! It's valuable to see the schema implemented in `squeal`. However, the database library is not decided yet. And the schema will be implemented later by @rashadg1030.

Issue #20, opened 2019-03-14
GitHub API Query Functions


rashadg1030:
I've made a fork of the project and have been testing the GitHub API query functions located in issue-wanted/src/IssueWanted/Search.hs. None of the functions return errors which is great, but I'm not sure if they are returning the right results. For example, the function `fetchHaskellReposGFI`, which returns all Haskell repositories with "good-first-issue" labels, gives a result count of 124. The function `fetchGoodFirstIssue`, which is supposed to return all issues with Haskell language and label "good-first-issue", only gives a result count of 8. This seems odd to me, but I'm not sure. There may be a problem with the query strings passed into `searchRepos` or `searchIssues`, but I can't be sure until I look more into the GitHub API documentation. I just wanted to get someone else's opinion on this issue.

ChShersh:
@rashadg1030 I think it's better to create minimal example and open issue for GitHub repo. But the chances is that you see lower number of results due to pagination. Though, number 8 looks too small for that. So it is still a good idea to ask about the behavior under `github` package:

* https://github.com/phadej/github/issues

rashadg1030:
@ChShersh Okay that makes sense. I'll do some more research. Thank you.

PR #22, opened 2019-03-15
[#20] Fix bugs in GitHub API query functions

rashadg1030:
I think I have fixed the GitHub API query functions. I have changed the query strings passed into the `searchIssues` functions to the correct ones. Now, instead of just returning 8 results each, they both seem to return the correct amount of results. I have tested the query strings using curl and Postman.


ChShersh:
@rashadg1030 Thanks for your contribution! I appreciate that you put effort into contributing. But no rush here 🙂 It's too early to work on the project. More details about the plan for the `issue-wanted` will be clarified closer to the project start.

PR #24, opened 2019-04-06
Search repos using the label: "low-hanging-fruit"

Infi-Knight:
Signed-off-by: Ravi Soni <soniravi829@gmail.com>


chshersh:
@Infi-Knight Thanks for your contribution! I think we will aim for a more general approach of handling labels instead of hardcoding some functions.

PR #31, opened 2019-05-24
[#14] Change module prefix to IW

### LINKED ISSUE #14: Choose shorter name for `IssueWanted` prefix
All modules currently have `IssueWanted` prefix in the library. Let's choose shorter prefix until it's not to late :slightly_smiling_face:

rashadg1030:
Resolves #14 

In order to change the module names and get the project to build I
also needed to rename the exposed-modules in the issue-wanted.cabal
file.

### DIFF
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -19,8 +19,8 @@ tested-with:         GHC == 8.4.3
 library
   hs-source-dirs:      src
   exposed-modules:     Prelude
-                       IssueWanted
-                         IssueWanted.Search
+                       IW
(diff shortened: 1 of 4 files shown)


rashadg1030:
> @rashadg1030 Nice! Yes, you indeed need to change the name of the modules in the `.cabal` file as well.
> 
> In terms of workflow: it's better if you write `Resolves #14` in the PR body message so issue is closed automatically when the PR is merged. If you use `hit commit` then it will add this line automatically to your commit message.

@chshersh Got it!

Issue #25, opened 2019-05-22
Upgrade to GHC-8.6.5


chshersh:
- [ ] Upgrade cabal file (`base` dependency, use `cabal-version: 2.4`, use common stanzas)
- [ ] Upgrade `stack.yaml` (bump up lts, libraries versions)
- [ ] Update `.travis.yml` file

As a reference, you can see changes in the [`three-layer`](https://github.com/holmusk/three-layer) repository.

PR #34, opened 2019-05-25
[#25] Update cabal and stack.yaml files

rashadg1030:
In order to completely resolve issue #25, I also need to update the `travis.yml` file but I'm not sure how to go about this. The `three-layer` example repository seems to use `circleci`, so I can't really use it as a reference. Should we use circle-ci instead? Any suggestions?

### DIFF
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -1,7 +1,7 @@
 cabal-version:       2.4
 name:                issue-wanted
 version:             0.0.0
-description:         Please see the README on Github at <https://github.com/kowainik/issue-wanted#readme>
+description:         Web application to help beginners to start contributing into Haskell projects
 synopsis:            Web application to help beginners to start contributing into Haskell projects
 homepage:            https://github.com/kowainik/issue-wanted#readme
 bug-reports:         https://github.com/kowainik/issue-wanted/issues
@@ -23,7 +23,6 @@ source-repository head
 common common-options
   build-depends:       base-noprelude ^>= 4.12.0.0
                      , relude ^>= 0.5.0
-                     , github >= 0.21
 
   ghc-options:         -Wall
                        -Wincomplete-uni-patterns
@@ -49,6 +48,7 @@ library
   exposed-modules:     Prelude
                        IW
                        IW.Search
+  build-depends:       github == 0.20 
 
 executable issue-wanted
   import:              common-options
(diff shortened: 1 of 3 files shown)


rashadg1030:
I fixed a typo in the commit message.

issue-wanted.cabal
```diff
@@ -13,14 +13,17 @@ copyright:           2018 Kowainik
...
-                       IW
-                       IW.Search
+source-repository head
+  type:                git
+  location:            https://github.com/kowainik/issue-wanted.git
+
+common common-options
+  build-depends:       base-noprelude ^>= 4.12.0.0
+                     , relude ^>= 0.5.0
+                     , github >= 0.21
```
vrom911: At the moment not all stanzas using `github` dependency, so it's better to add only dependencies that are used everywhere.
And also, the upper boundary is not specified. It is much better to have all the boundaries specified for the libraries.

By the way, `github-0.21` is very broken, so I would suggest using an older version. We are using `github-0.20` in another project.

stack.yaml
```diff
@@ -1,5 +1,6 @@
-resolver: lts-12.9
+resolver: lts-13.20
```
vrom911: `lts-13.21` is already out! :slightly_smiling_face:

issue-wanted.cabal
```diff
@@ -1,9 +1,9 @@
-cabal-version:       2.2
+cabal-version:       2.4
 name:                issue-wanted
 version:             0.0.0
-description:         Web application to help beginners to start contributing into Haskell projects
+description:         Please see the README on Github at <https://github.com/kowainik/issue-wanted#readme>
```
vrom911: No need to change the description on the link to README. If you feel that it's not very informative we could change it to something more useful, what do you think?

issue-wanted.cabal
rashadg1030: Okay got it. I will remove `github` from the common stanza. Should the boundary be like 
`github >= 0.20 && < 0.21` or just `github == 0.20`?

issue-wanted.cabal
rashadg1030: Got it. I think the old description is fine.

stack.yaml
rashadg1030: Nice! I’ll change it

issue-wanted.cabal
rashadg1030: I used github == 0.20

vrom911: Good! Until the new non-broken version is out we can use this one. We will keep an eye on this 🙂

Issue #5, opened 2018-06-01
Describe all API endpoints we need from GitHub


chshersh:
Basically we need:

- [x] List of all Haskell repositories
```
curl -H 'Accept: application/vnd.github.preview.text-match+json' https://api.github.com/search/repositories?q=language:haskell&order=desc
```

- [ ] For each repository: list of issues by given label (`help wanted`, `good first issue`)
    + **TODO**: insert Rest API method here
- [ ] For each user: each PR for this user
    + **TODO**: insert Rest API method here
- [ ] For each PR: which issue it closes
    + **TODO**: insert Rest API method here

willbasky:
> For each user: each PR for this user

`For each user` on our site or at GitHub? Could it described more?

ChShersh:
@willbasky In this context `For each X` means _provide function `X -> Smth`_. We need to know how to fetch information about user. And then we can easily fetch this information for multiple users. But what is the set of users we're interested in – is another task.

rashadg1030:
@ChShersh We have a working GitHub query function `fetchAllHaskellRepos` that returns a vector of repos. Would it be fine to write a function that takes a repo and returns a list/vector of issues with the `help wanted`/`good first issue` labels? `Repo -> [Issues]` for example.

ChShersh:
@rashadg1030 I think it's too early to think about this. Also see issue #21, some parts of the architecture might change before project starts.

rashadg1030:
@ChShersh Alright, thank you. I will look more into issue #21. I think it is a good idea.

rashadg1030:
@chshersh I would like to discuss more the type of endpoints we will be exposing and the data that they represent. Maybe we could start listing endpoints that we think issue-wanted should have?

rashadg1030:
##For each repository: list of issues by given label (help wanted, good first issue):

1. https://api.github.com/repos/:repo_owner/:repo_name/issues?labels=help-wanted
2. https://api.github.com/repos/:repo_owner/:repo_name/issues?labels=good-first-issue

##For each user: each PR for this user:

Interesting Stack Overflow comment I found regarding this issue

`First you have to realize that you must authenticate using either Basic Authentication or a token. Next you have to realize that there is no simple way to do this so you will have to be clever.
[...]

PR #36, opened 2019-06-01
 [#5] Add new function fetchIssuesWithLabels

rashadg1030:
Added a new function fetchIssuesWithLabels that takes a list of labels and returns a list of open, Haskell issues with those labels. I've also modified the functions fetchHelpWanted and fetchGoodFirstIssue to use
this new function.


src/IW/Search.hs
```diff
@@ -18,10 +17,17 @@ fetchHaskellReposHW = searchRepos "language:haskell help-wanted-issues:>0"
 fetchHaskellReposGFI :: IO (Either Error (SearchResult Repo))
 fetchHaskellReposGFI = searchRepos "language:haskell good-first-issues:>0"
 
--- | Fetch all issues with Haskell language and label "help-wanted"
+-- | Fetch all open issues with Haskell language and the labels passed in to the function
+fetchOpenIssuesWithLabels :: [Text] -> IO (Either Error (SearchResult Issue))
+fetchOpenIssuesWithLabels labels = searchIssues $ "language:haskell is:open " <> labelQualifiers
+  where
+    labelQualifiers :: Text
+    labelQualifiers = unwords $ (\x -> " label:\"" <> x <> "\"") <$> labels
```
vrom911:
```suggestion
    labelQualifiers = concatMap (\x -> " label:\"" <> x <> "\" ") labels   
```
I guess this could be written this way.

src/IW/Search.hs
```diff
@@ -18,10 +18,17 @@ fetchHaskellReposHW = searchRepos "language:haskell help-wanted-issues:>0"
 fetchHaskellReposGFI :: IO (Either Error (SearchResult Repo))
 fetchHaskellReposGFI = searchRepos "language:haskell good-first-issues:>0"
 
--- | Fetch all issues with Haskell language and label "help-wanted"
+-- | Fetch all open issues with Haskell language and the labels passed in to the function
+fetchOpenIssuesWithLabels :: [Text] -> IO (Either Error (SearchResult Issue))
+fetchOpenIssuesWithLabels labels = searchIssues $ "language:haskell is:open " <> labelQualifiers
+  where
+    labelQualifiers :: Text
+    labelQualifiers = mconcatMap (\x -> " label:\"" <> x <> "\" ") labels
```
chshersh: Hmm, interesting :thinking: I had an impression that `concatMap` should work here...
However, you don't need to depend on the `extra` package. You can just use `foldMap` instead of `mconcatMap`.

rashadg1030: Okay I can change. concatMap didn’t work because it’s Text. If it was String I think it would work though. I’ll checkout the foldMap function. Thank you!

PR #38, opened 2019-06-04
[#29] Change file structure

### LINKED ISSUE #29: Bring `three-layer` architecture to issue-wanted
See the [`three-layer`](https://github.com/holmusk/three-layer) repository for example. 

Specifically, we need to bring the following things:

- [x] Our custom monad with the environment (see `Lib/App` directory)
- [x] Error data types with helper functions (see `Lib/App/Error.hs` module)
- [x] `Effects.Log` module with logging
- [x] TOML configuration
- [x] Makefile for running commands smoothly

rashadg1030:
Can anybody confirm that this is the correct file structure? Thank you! I think I've introduced most of the things we're gonna need for now. Let me know if I'm missing anything.

Resolves #29


src/IW/App.hs
```diff
@@ -0,0 +1,9 @@
+module IW.App
+( module IW.App.Env
+, module IW.App.Error
```
chshersh: According to our style guide, `(` and `,` should be aligned under module name with 7 spaces. Like this:

```haskell
module IW.App
       ( module IW.App.Env
       , ...
       ) where
```

src/IW/App/Error.hs
```diff
@@ -0,0 +1,193 @@
+{-# LANGUAGE ConstraintKinds #-}
```
chshersh: `ConstraintKinds` can be put into `default-extensions`

src/IW/App/Error.hs
```diff
@@ -0,0 +1,193 @@
...
+    -}
+    | MissingHeader HeaderName
+    {- | An authentication header that was required was provided but not in a
+    format that the server can understand
+    -}
+    | HeaderDecodeError Text
+    -- | Data base specific errors
+    | DbError Text
+    -- | Limits on the multi-request are overflowed.
+    | LimitError
```
chshersh: As discussed, we probably won't need `LimitError`. But there is no harm in having it :slightly_smiling_face:

rashadg1030: Oh yeah, I forgot about that. Got it.

src/IW/App/Error.hs
rashadg1030: Got it

src/IW/App.hs
rashadg1030: Yep that's wrong. Will fix it. Thank you.

src/IW/App/Error.hs
rashadg1030: I'll just remove `LimitError` for now and bring it back later if needed.

issue-wanted.cabal
```diff
@@ -34,23 +34,82 @@ common common-options
...
+                     , github == 0.20
+                     , http-api-data ^>= 0.4
+                     , http-types ^>= 0.12
+                     , jwt ^>= 0.9.0
+                     , lens ^>= 4.17
+                     , mtl ^>= 2.2.2
+                     , postgresql-simple ^>= 0.6.1
+                     , proto-lens ^>= 0.5.0.0
+                     , proto-lens-protoc ^>= 0.5.0.0
+                     , proto-lens-runtime ^>= 0.5.0.0
```
chshersh: We won't need `proto-lens` libraries, you can remove them as well

issue-wanted.cabal
```diff
@@ -34,23 +34,82 @@ common common-options
...
+                           IW.Db.Schema
+                      
+                         -- Effects
+                         IW.Effects.Log  
+                         
+                         -- Server
+                         IW.Server
+                           IW.Server.Search
+  build-depends:       aeson >= 1.4
+                     , bcrypt ^>= 0.0.11
```
chshersh: We also don't need `bcrypt`. I would propose to try to remove redundant dependencies to make the build faster. Let's not introduce libraries in our `build-depends` unless we use them.

For now it's okay to keep `servant` and `postgresql-simple` dependencies and reexports from `Prelude` for those libraries. Those two will be brought here very soon, so it's okay to have them for now. But other redudant libraries look redundant indeed (and don't forget to remove them from `stack.yaml` as well).

issue-wanted.cabal
```diff
@@ -34,23 +34,82 @@ common common-options
...
+                     , servant-proto-lens ^>= 0.1.0.3
+                     , servant-server ^>= 0.16
+                     , text
+                     , time >= 1.8 && < 1.10
+                     , tomland ^>= 1.0.0
+                     , unordered-containers
+                     , warp ^>= 3.2
 
-executable issue-wanted
+executable issue-wanted-exe
```
chshersh: Is there any reason to add `-exe` suffix here? I think it can be just `issue-wanted` for now (also, don't forget to change `Makefile` after renaming).

src/IW/App/Monad.hs
```diff
@@ -0,0 +1,48 @@
+module IW.App.Monad
+    ( -- * Application monad
+      App (..)
+    , AppEnv
```
chshersh: Reexports are not following the style guide here as well.

src/Prelude.hs
```diff
@@ -1,7 +1,48 @@
--- | Uses [relude](https://github.com/kowainik/relude) as default Prelude.
+{-# OPTIONS -fno-warn-orphans #-}
```
chshersh: This option can be removed from this file, we don't have orphan instances here for now.

stack.yaml
```diff
@@ -1,5 +1,19 @@
...
+  - jwt-0.9.0
+  - proto-lens-0.5.0.0
+  - proto-lens-protoc-0.5.0.0
+  - proto-lens-runtime-0.5.0.0
+  - relude-0.5.0
+  - servant-0.16.0.1
+  - servant-server-0.16
+  - servant-proto-lens-0.1.0.3
+  - tomland-1.0.0
+  - typerep-map-0.3.2
```
chshersh: If some library is not specified in `build-depends` then it means that it's used implicitly by some other libraries. But if stackage snapshot doesn't contain those libraries, then we need to add them to `extra-deps`. When this happens, we usually add comment after dependency why it's needed. This makes future cleanup easier. So, for example, all `co-log`-related dependencies in `extra-deps` should look like this (modulo only relevant libraries):

```yaml
# logging
- co-log-0.3.0.0
- ansi-terminal-0.9.1        # for co-log
- chronos-1.0.5              # for co-log
- co-log-core-0.2.0.0        # for co-log
- torsor-0.1                 # for co-log
- typerep-map-0.3.2          # for co-log
```

issue-wanted.cabal
rashadg1030: I got the idea from here: https://github.com/Holmusk/three-layer/blob/eedbed6d1795eefe7d23779ee83ee7e0885bd9b1/three-layer.cabal#L151-L157

I will change it to just `issue-wanted` for now

Makefile
```diff
@@ -0,0 +1,8 @@
+ide:
+	ghcid --command "stack ghci --ghci-options=-fno-code --main-is issue-wanted:lib issue-wanted:exe:issue-wanted-exe issue-wanted:test:issue-wanted-test"
```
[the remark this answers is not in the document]
rashadg1030: Got it.

src/IW/App/Monad.hs
rashadg1030: Will fix!

src/Prelude.hs
rashadg1030: Got it.

stack.yaml
rashadg1030: Fixed. Let me know if I did it correctly.

Issue #30, opened 2019-05-22
Describe SQL schema for storing data


chshersh:
What needs to be done here:

- [x] Schema should be written in raw SQL files in the `sql/` directory in the project root. This directory should contain two files now: `schema.sql` with the schema and `drop.sql` with removing the schema (useful for testing)
- [x] `IW.Db.Schema` module with helper functions (see `three-layer` for the example)

rashadg1030:
I want to work on the schema but I'm not sure what the shape of our data is. Any suggestions? I have some ideas that I will post later.

chshersh:
@rashadg1030 You can start from seeing what GitHub API provides and think of possible tables and columns that can be relevant. So you can start with something and improve the schema later during the work when we realise that we need something more.

rashadg1030:
@chshersh Okay so I'm thinking we have two tables: `repos` and `issues`. I think issues should be related to repos using a foreign key. Each issue will have a repo id as a foreign key. Or should we just keep the two tables separate and ignore the relationship. Which is better?

chshersh:
@rashadg1030 I think your proposal is good :+1: The only difficult question here is to how store labels for issues and categories for repos so they can be filtered efficiently. Note, that issues may change their labels, some labels can be renamed and we want to keep up-to-date information in our DB. So there should be some way to update this information.

rashadg1030:
@chshersh The best solution I can think of for storing labels with issues is storing the `label_url` for each issue in the DB. I just need to add a `label_url` field to the issues table. The server can use the `label_url` to get all labels for an issue. Or that can be done by the frontend. The frontend could ask the server for a `label_url`, make a request for labels using the url, and then display it.

rashadg1030:
@chshersh For repository categories I'm not sure. By repository "category" do you mean GitHub [topics](https://help.github.com/en/articles/classifying-your-repository-with-topics)? Do you mean the hackage category the package belongs to? Or will we have our own categories for repositories somehow?

rashadg1030:
@chshersh I just read this article on handling many-to-many relationships in a DB. We could uses a `join table` to associate `issues` with `labels`. [Link](https://dzone.com/articles/how-to-handle-a-many-to-many-relationship-in-datab)

PR #40, opened 2019-06-10
[#30] Describe SQL schema

rashadg1030:


rashadg1030:
I think this pull request is pretty much done. I'm just concerned about the SQL naming convention and the constraints I added to the tables but I tested it out in the SQL repl and it seems to be fine. Let me know if I'm missing anything.

issue-wanted.cabal
```diff
@@ -101,7 +101,7 @@ library
                      , unordered-containers
                      , warp ^>= 3.2
 
-executable issue-wanted
+executable issue-wanted-exe
```
chshersh: I think this is a minor mistake introduce during conflict resolution

sql/schema.sql
```diff
@@ -0,0 +1,89 @@
...
+-- BASE TABLES --
+-----------------
+
+CREATE TABLE IF NOT EXISTS repos
+( id               INT  NOT NULL
+, repo_name        TEXT NOT NULL
+, repo_desc        TEXT NOT NULL
+, repo_url         TEXT NOT NULL -- The repo html_url
+, open_issue_count INT  NOT NULL
+, fork_count       INT  NOT NULL
```
chshersh: We probably won't need everything from here, but there is no harm in having more fields, the schema can be always adjusted later.

issue-wanted.cabal
rashadg1030: Ooops! Thank you.

sql/schema.sql
rashadg1030: Got it.

PR #44, opened 2019-06-16
[#26] Implement a servant endpoint

rashadg1030:
Added a single endpoint for getting issues by id. Seed the DB with `sql/seed.sql` (one example issue), execute the application, and then use this endpoint: http://localhost:8081/issues/1 
This should return an issue with issueId 1.


rashadg1030:
Resolved conflicts

src/IW/Core/Issue.hs
```diff
@@ -0,0 +1,17 @@
...
+data Issue = Issue
+    { issueId     :: Int
+    , issueNumber :: Int
+    , issueTitle  :: Text
+    , issueBody   :: Text
+    , issueUrl    :: Text
+    , issueRepoId :: Int
+    } deriving stock (Generic, Show, Eq)
+      deriving anyclass (FromRow)
+      deriving (FromJSON, ToJSON)
```
vrom911: These two can also be derived as `anyclass`, so two lines can be merged together.

src/IW/Server/Types.hs
```diff
@@ -0,0 +1,15 @@
...
+       ) where
+
+import Servant.API.Generic (ToServantApi)
+import Servant.Server.Generic (AsServerT)
+
+import IW.App (App)
+
+
+type AppServer = AsServerT App
+type ToApi (site :: * -> *) = ToServantApi site
```
chshersh: Let's write here `Type -> Type` as well instead of `* -> *` for consistency.

src/IW/Server/Issue.hs
```diff
@@ -0,0 +1,37 @@
...
+issueServer :: IssueSite AppServer
+issueServer = IssueSite
+    { issueByIdRoute  = issueByIdHandler
+    }
+
+issueByIdHandler
+    :: ( WithDb env m
+       , WithError m
+       )
+    => Int
```
chshersh: When we bring `Id` from `three-layer`, this argument should be changed to `Id Issue`.

src/IW.hs
```diff
@@ -1,2 +1,32 @@
...
+mkAppEnv Config{..} = do
+    -- IO configuration
+    envDbPool <- initialisePool cDbCredentials
+
+    -- pure configuration
+    let envLogAction = mainLogAction cLogSeverity
+    pure Env{..}
+
+runServer :: AppEnv -> IO ()
+runServer env@Env{..} = run 8081 application
```
chshersh: This is just a convention, but for running backend we are usually using port `8080`. `8081` for additional server

src/IW/Core/Issue.hs
rashadg1030: Got it.

src/IW/Server/Types.hs
rashadg1030: Oh yeah got it.

src/IW/Server/Issue.hs
rashadg1030: Got it.

src/IW.hs
rashadg1030: Got it. Changing it now.

rashadg1030:
@chshersh @vrom911 I made the fixes you suggested! I think I'm going to open an issue for introducing the `Id` type from `three-layer`.

Issue #46, opened 2019-06-17
Issue-wanted endpoint URLs


rashadg1030:
What should the URLs for `issue-wanted` look like? So far, we have one endpoint:
`~/issues/:issueId`

Some other ones I think we need:

```
~/issues/         -- returns all issues
~/issues/:label   -- returns all issues with the given label 
~/repos/          -- returns all repos
~/repos/:repoId   -- returns a repo with the Id
~/repos/:category -- returns all repos with the given category
```

Any more suggestions?

chshersh:
For now we are interested only in the following endpoints:

1. Get all issues (sorted by some default parameters)
2. Get all issues that have a label or category from the given list (filtering).

All queries should also take limit with default value (like 100) (we don't want to give 150K issues each time). Pagination will be implemented later.

rashadg1030:
@chshersh Sort parameter examples:
`~/issues?sort=title:asc` Sort by issue title ascending order
`~/issues?sort=id:desc`  Sort by issue id descending order
`~/issued?sort=id:asc&label=good first issue` Sort by ascending issue id and has label good first issue

Is there any more that will be helpful?

rashadg1030:
My next goal is to add a limit query parameter maybe? How will we do pagination?

PR #47, opened 2019-06-18
[#46] Add more endpoints to issue-wanted API

rashadg1030:
Need help with how to handle query parameters in servant route. I think I have a solution, but wondering if there is a better way. Look at `src/IW/Server/Issue.hs`.


rashadg1030:
Fleshed out the issues endpoint. It takes `sort`, `label`, and `desc` query parameters. For example:
http://localhost:8081/issues?sort=title&label=help%20wanted&desc
I'm sure there is a better way to handle the sorting, but this is the best way I could come up. A lot of the sorting is done with Haskell. I wanted to use `ORDER BY` and stuff like that on that on the SQL side (it's probably more performant), but I didn't couldn't come up with a solution.

src/IW/Db/Issue.hs
```diff
@@ -0,0 +1,35 @@
...
+       , getIssuesByLabel
+       ) where
+
+import IW.App (WithError)
+import IW.Core.Issue (Issue)
+import IW.Db.Functions (WithDb, asSingleRow, query, queryRaw)
+
+getIssues :: (WithDb env m, WithError m) => m [Issue]
+getIssues = queryRaw [sql|
+    SELECT *
```
vrom911: `SELECT *` is never a good idea. Let's be implicit instead :slightly_smiling_face:

src/IW/Server/Issue.hs
```diff
@@ -0,0 +1,83 @@
...
+       , issueByIdHandler
+       , issuesHandler
+       ) where
+
+import IW.App.Error (WithError)
+import IW.Core.Issue (Issue (..))
+import IW.Db (WithDb, getIssues, getIssueById, getIssuesByLabel)
+import IW.Server.Types (AppServer, ToApi)
+
+data Sort = Id | Title
```
vrom911: Better to have all data types in `Core` modules for consistency.

src/IW/Server/Issue.hs
```diff
@@ -0,0 +1,83 @@
...
+
+data IssueSite route = IssueSite
+    { -- | Fetch issue by issue ID
+      issueByIdRoute :: route
+        :- "issues"
+        :> Capture "id" Int
+        :> Get '[JSON] Issue
+    , issuesRoute :: route
+        :- "issues"
+        :> QueryParam "sort" Sort
```
vrom911: I'm not sure if it is ok to have this as a query parameter. I think it is not an optimal solution and could be too slow. What about the possibility of having sorting on the frontend side?

src/IW/Server/Issue.hs
```diff
@@ -0,0 +1,83 @@
...
+issuesHandler 
+    :: ( WithDb env m
+       , WithError m
+       )
+    => Maybe Sort
+    -> Bool
+    -> Maybe Text
+    -> m [Issue]
+issuesHandler sortParam isDesc labelParam = do 
+  issues       <- case labelParam of
```
vrom911: This is a bit of a strange layout, probably, we could skip these spaces? And also we are using 4 spaces indentation by style guide

src/IW/Server/Issue.hs
```diff
@@ -0,0 +1,83 @@
...
+    sortById :: [Issue] -> [Issue]
+    sortById = sortOn issueId
+
+    sortByTitle :: [Issue] -> [Issue]
+    sortByTitle = sortOn issueTitle
+
+    sortIssues :: Maybe Sort -> [Issue] -> [Issue]
+    sortIssues Nothing      = id
+    sortIssues (Just sort') = case sort' of
+                                Id    -> sortById
```
vrom911: We have 4-space indentation policy by our style guide

src/IW/Server/Issue.hs
chshersh: @vrom911 I agree. Calling backend for sorting is a bit overkill and will result in slow performance. I propose to think carefully about what we want to do on frontend and backend. Here are my thoughts:

1. By default frontend initially frontend asks for 100 latest created issues.
2. Sorting is done purely on frontend. I think for now it makes sense to sort only by time of creation.
3. We need to discuss how we want to filter issues. We might want to filter by:
    * Zero or more labels
    * Zero or more categories

Filtering should be done on the backend, because if frontend knows only 100 recent all issues, it can't know 100 recent issues with the label `help wanted`

sql/schema.sql
```diff
@@ -6,31 +6,31 @@
...
-, repo_desc        TEXT NOT NULL
-, repo_url         TEXT NOT NULL -- The repo html_url
-, open_issue_count INT  NOT NULL
-, fork_count       INT  NOT NULL
+( id         SERIAL       
+, repo_name  TEXT   NOT NULL
+, repo_owner TEXT   NOT NULL
+, repo_desc  TEXT
+-- The repo's html_url
+, repo_url   TEXT   NOT NULL
```
chshersh: I think we don't need to store `repo_url`. It can be constructed from owner and name

sql/schema.sql
```diff
@@ -6,31 +6,31 @@
...
-( id               INT  NOT NULL
-, repo_name        TEXT NOT NULL
-, repo_desc        TEXT NOT NULL
-, repo_url         TEXT NOT NULL -- The repo html_url
-, open_issue_count INT  NOT NULL
-, fork_count       INT  NOT NULL
+( id         SERIAL       
+, repo_name  TEXT   NOT NULL
+, repo_owner TEXT   NOT NULL
+, repo_desc  TEXT
```
chshersh: Let's not prefix columns with the `repo_` or table name. It is a little bit redundant because during SQL queries we can always refer to columns by fully qualified name like `repos.name` or `repos.owner` in case of ambiguity. So in tables, in general, it's better to have ordinary names.

sql/schema.sql
```diff
@@ -6,31 +6,31 @@
...
+, issue_title  TEXT   NOT NULL
+, issue_body   TEXT
+, issue_url    TEXT   NOT NULL
+, repo_id      INT    NOT NULL
 );
 
 CREATE TABLE IF NOT EXISTS categories
-( id            INT  NOT NULL
-, category_name TEXT NOT NULL
+( id            SERIAL
```
chshersh: Just my personal taste, by I prefer to have `PRIMARY KEY` near `SERIAL`. Constraints and foreign keys can be big and they change often, so it's better to specify them separately via `ALTER` statements. But changing the primary key will result in changing everything, so it's okay to specify it directly in the table.

src/IW/Db/Issue.hs
```diff
@@ -0,0 +1,35 @@
...
+    FROM issues
+|]
+
+getIssueById :: (WithDb env m, WithError m) => Int -> m Issue
```
chshersh: I think you can create issue about introducing `newtype Id` with phantom type parameter and use it instead of `Int` where you need `Id` in a separate PR (after this PR is merged).

src/IW/Server/Issue.hs
```diff
@@ -0,0 +1,83 @@
...
+    -> Maybe Text
+    -> m [Issue]
+issuesHandler sortParam isDesc labelParam = do 
+  issues       <- case labelParam of
+                    Nothing -> getIssues
+                    Just p  -> getIssuesByLabel p
+  return $ sortDesc isDesc . sortIssues sortParam $ issues
+  where
+    sortById :: [Issue] -> [Issue]
+    sortById = sortOn issueId
```
chshersh: We decided to do sorting on frontend. But just for future reference:

1. It's better to do sorting in the SQL query directly rather than in memory. Databases can sort faster and more efficiently, and linked lists are not the best data structure for sorting.
2. If you want to sort by some field, it's better to use `sortWith` instead of `sortOn`. `sortOn` creates tuples and sort tuples, so `sortOn` is more efficient if you want to sort by some expensive function, like, length for lists. But if you're just sorting by field then it's better to use `sortWith`.

chshersh:
The ideas look good :+1: But I think that we need to discuss a lot :slightly_smiling_face:

src/IW/Db/Issue.hs
rashadg1030: Got it.

src/IW/Server/Issue.hs
rashadg1030: Got it. I was kinda just using this as a temporary prototype to see if this was the right path to go down.

src/IW/Server/Issue.hs
rashadg1030: Sounds good. I thought the same, and wanted to try and do sorting at the database level with SQL, but didn't know how to implement it. I was thinking of creating a whole bunch of different SQL queries for each type of sorting situation, but that wouldn't be fun lol

src/IW/Server/Issue.hs
rashadg1030: Yes, my bad I will fix this.

src/IW/Server/Issue.hs
rashadg1030: My bad again will fix.

sql/schema.sql
rashadg1030: Got it.

sql/schema.sql
rashadg1030: Got it.

sql/schema.sql
rashadg1030: Okay cool that makes sense. Will fix it.

src/IW/Db/Issue.hs
rashadg1030: I will do that.

src/IW/Server/Issue.hs
rashadg1030: Got it! Thanks for the info.

rashadg1030:
@chshersh @vrom911 I have made changes and I think this PR is good to go. Thanks for the tips!

rashadg1030:
@chshersh @vrom911  Okay I resolved any conflicts. Should be good to go.

rashadg1030:
@chshersh This branch is ready to merge but it says I need to resolve one more change request. I think it's referring to the `Id` type, but I have another for this.

PR #49, opened 2019-06-22
[#48] Introduce Id type with phantom type parameter

rashadg1030:


src/IW/Core/Repo.hs
```diff
@@ -0,0 +1,9 @@
+{-# LANGUAGE DeriveAnyClass #-}
+
+module IW.Core.Repo
+       ( Repo (..)
+       ) where
+        
+
+-- | Data type representing a GitHub repository.
+data Repo = Repo
```
vrom911: This data type doesn't look right. Probably if it should only be used as a phantom param then it should not have any constructors at all?

rashadg1030: Yes, thank you!

Issue #54, opened 2019-06-26
Update the schema for more efficiency and simplicity


chshersh:
I did some research on how to structure the database for our use case better. We need to do some adjustments to make it work fast. At the same time, it will make our lives easier.

1. Foreign key in the `issues` table should be not for id but for `repo.name`. It's completely okay to have textual foreign keys. Just add an INDEX for `repos.name` column (later we can see what columns we are using so we can add more indexes for performance).
    * https://www.postgresql.org/docs/9.1/sql-createindex.html
2. Remove tables `categories`, `labels`, `repos_categories` and `issues_labels`. Instead, we will store them as a separate column using [PostgreSQL arrays](https://www.postgresql.org/docs/9.1/arrays.html). This approach is much much faster than joining tables. And such arrays support all operations that we need for filtering. This also means that updating labels is now a simple task: just write new labels in place for that array.

rashadg1030:
Got it. I will work on this and ask if I have any questions. I've never heard of PostgreSQL arrays before but it seems useful.

rashadg1030:
Also one question. For the `Issue` instead of `repoId :: Id Repo` should we have a type `repoName :: Name Repo` or something like that?

rashadg1030:
Another question. I can't make `repo.name` a foreign key unless I add a `UNIQUE` constraint to it. GitHub repository names aren't unique right?

chshersh:
> For the Issue record, instead of repoId :: Id Repo should we have a type repoName :: Name Repo or something like that?

We will have `Id` in almost every data type. But I don't think we will have many names. Once we have `Repo` data type with fields, we can introduce our own `Name` newtype similar to `Id`. But for now `Text` should work.

> I can't make repo.name a foreign key unless I add a UNIQUE constraint to it. GitHub repository names aren't unique right?

Okay, in that case, let's add `owner` column to the `issues` table and have a foreign key for pair. See how it's done here:

https://stackoverflow.com/questions/25564036/is-it-possible-to-create-a-foreign-key-pair-sql

rashadg1030:
Got it👍🏽

PR #55, opened 2019-06-27
[#54] Update DB

rashadg1030:
```
ERROR:  there is no unique constraint matching given keys for referenced table "repos"
```
I'm still getting this error after adding the constraint:

```
ALTER TABLE ONLY issues
  ADD CONSTRAINT fk_repos FOREIGN KEY (owner, repo_name) 
  REFERENCES repos (owner, name) ON DELETE CASCADE;
```
Gonna try to fix it, but let me know if there is a good way to do this.


rashadg1030:
I'm not sure if this is possible, but I think for the repo table we can make the `owner` and `repo` the primary key. Can a primary key refer to two different columns? I'll try this.

rashadg1030:
Yes, it's possible. I propose the above idea. If so, does that mean we don't even need `Id`, or is it good to have that for convenience?

chshersh:
@rashadg1030 You can add UNIQUE constraint for pair as well. See how it's done here:

https://stackoverflow.com/questions/14221775/in-postgresql-force-unique-on-combination-of-two-columns

Let's try not to have `name` and `owner` be the primary key. It's much faster if the primary key will be just integer.

Also, I see that this PR is based on your sync worker. And Circle CI is failing because you don't have config from `master`. Could. you, please base this PR on top of `master` and not sync worker? Sync worker is a huge part and probably should be cut into multiple small pieces. We need to merge schema updates first before starting to work on a sync worker, not vice versa.

rashadg1030:
Got it. Will do that. Thank you for the link.

PR #64, opened 2019-06-27
[#57] Update Issue type with FromRow instance using PGArray

### LINKED ISSUE #57: Update Issue type

rashadg1030:
Resolves #57


rashadg1030:
Adding `ToRow` instance next.

src/IW/Core/Issue.hs
```diff
@@ -5,16 +5,33 @@ module IW.Core.Issue
...
-    , issueUrl    :: Text
-    , issueRepoId :: Id Repo
+    { issueId       :: Id Issue
+    , issueNumber   :: Int
+    , issueTitle    :: Text
+    , issueBody     :: Text
+    , issueUrl      :: Text
+    , issueOwner    :: Text
+    , issueRepoName :: Text
+    , issueLabels   :: PGArray' Text
```
chshersh: I propose to just use `PGArray Text` without JSON instances. We haven't decided on the frontend framework yet, so the solution for JSON serialisation will depend on our choice. And that's why let's try to not complicate the code at the moment.

rashadg1030: Got it

rashadg1030: If I don't have a `ToJSON` for `Issue` it won't compile.
```
No instance for (ToJSON IW.Core.Issue.Issue)
        arising from a use of ‘hoistServer’
    • In the expression:
        hoistServer
          (Proxy @API) (runAppAsHandler env) (toServant issueServer)
      In an equation for ‘server’:
          server env
            = hoistServer
                (Proxy @API) (runAppAsHandler env) (toServant issueServer)
   |
17 | server env = hoistServer (Proxy @API) (runAppAsHandler env) (toServant issueServer)
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rashadg1030:
I should probably move the `PGArray'` newtype and it's instances to it's own module in `IW.Core`.

src/IW/Core/PGArray'.hs
```diff
@@ -0,0 +1,22 @@
+{-# LANGUAGE DeriveAnyClass #-}
+
+module IW.Core.PGArray'
```
chshersh: `PGArray'` is a poor name because it doesn't reflect the purpose of this data type. I would propose to rename it to `SqlArray`. This name is more different from `PGArray` which signals that it has different semantics.

src/IW/Core/PGArray'.hs
```diff
@@ -0,0 +1,22 @@
...
+    ) where
+
+import Database.PostgreSQL.Simple.Types (PGArray (..))
+import Database.PostgreSQL.Simple.FromField (FieldParser, pgArrayFieldParser)
+import Database.PostgreSQL.Simple.ToField (Action (..))
+
+
+newtype PGArray' a = PGArray' { fromPGArray' :: [a] }
+    deriving stock (Generic, Show, Eq)
+    deriving anyclass (ToJSON)
```
chshersh: `Eq` instance for `newtype`s should always be derived using `newtype` deriving strategy.

`ToJSON` should be derived using `newtype` strategy in this particular use case.

src/IW/Core/PGArray'.hs
```diff
@@ -0,0 +1,22 @@
...
+
```
chshersh: I wonder if you can use `DerivingVia` to reduce some boilerplate. And write something like:

```haskell
    deriving (ToField, FromField) via PGArray a
```

instead of these two manual instances.

src/IW/Core/PGArray'.hs
```diff
@@ -0,0 +1,22 @@
...
+module IW.Core.PGArray'
+    ( PGArray' (..)
+    ) where
+
+import Database.PostgreSQL.Simple.Types (PGArray (..))
+import Database.PostgreSQL.Simple.FromField (FieldParser, pgArrayFieldParser)
+import Database.PostgreSQL.Simple.ToField (Action (..))
+
+
+newtype PGArray' a = PGArray' { fromPGArray' :: [a] }
```
chshersh: After you rename it to `SqlArray` the name of the field should be `unSqlArray` according to our style guide.

chshersh:
I want to point out that we have a bit difficult problem here. That's why this seems-easy PR takes more time and discussion. I never had this problem before when we need to use some `newtype` from one library and instance from another library but when data type doesn't have such instance. Introducing our own `newtype` and deriving all instances accordingly looks like a reasonable solution at the moment.

src/IW/Core/PGArray'.hs
rashadg1030: @chshersh That's funny I actually had it as `SqlArray` originally but change it back. I agree `SqlArray` is better,

src/IW/Core/PGArray'.hs
rashadg1030: Got it. Will fix.

src/IW/Core/PGArray'.hs
rashadg1030: I will try this.

src/IW/Core/PGArray'.hs
rashadg1030: Got it.

rashadg1030:
@chshersh Yeah haha this PR has been interesting. Gonna clean it up.

PR #65, opened 2019-06-27
[#63] Add functions for testing DB functions without running entire app

### LINKED ISSUE #63: Add DB functions that operate within IO monad instead of entire App monad for testing

rashadg1030:
Resolves #63


rashadg1030:
@chshersh Okay no worries. I didn't spend a lot of time. I just the functions might be useful, but in hindsight they are kind of redundant. I will change it.

rashadg1030:
@chshersh  I also didn't know that about GHCi. That's good to know that you can just create `env` once and use it throughout the session.

PR #66, opened 2019-06-27
[#51] Add instructions for testing database

### LINKED ISSUE #51: Write instructions on how to build and test

rashadg1030:
Resolves #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,17 @@ single location. The goals of `issue-wanted` are to make it easier for
 programmers of all skill levels to find Haskell projects to contribute to,
 increase the number of contributions to open-source Haskell projects, and
 encourage more programmers to become a part of the Haskell community.
+
+# Testing
+
+## Testing the Database
+
+For testing the database follow these steps:
+
+1. Open up the terminal and run `cd test/Db`
+
+2. Run the command `stack ghci` to get into the REPL
+
+3. Run the command `:l Test` to load the database testing module
+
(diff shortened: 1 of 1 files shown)


rashadg1030:
Let me know if the order is right. Do we need the DB part?

README.md
```diff
@@ -10,3 +10,57 @@ single location. The goals of `issue-wanted` are to make it easier for
...
+
+  Returns a single issue with the corresponding id.
+
+* `http://localhost:8080/issues?label=<label name>`
+
+  Returns a list of issues with the corresponding label.  
+
+Tip: If you stop running the server and then try to restart it, you may need to run the command `fuser -n tcp -k 8080` to free the port.
+
+## How to test DB query functions
```
chshersh: I think that section can be removed from now. Testing DB queries from GHCi is more like a developer workflow. We can have a separate section like `## For Developers`, but since PR with GHCi commands is not merged yet, let's not write about this. Also because something might change here.

README.md
```diff
@@ -10,3 +10,45 @@ single location. The goals of `issue-wanted` are to make it easier for
 programmers of all skill levels to find Haskell projects to contribute to,
 increase the number of contributions to open-source Haskell projects, and
 encourage more programmers to become a part of the Haskell community.
+
+# Testing
```
chshersh: Heh, I forgot to mention, that this `# Testing` header is also redundant because not everything here is about testing.

rashadg1030: Oh okay got it.

PR #68, opened 2019-06-28
[#61] Add functions for inserting issues

### LINKED ISSUE #61: Add functions for inserting issues into DB

rashadg1030:
Resolves #61 

Depends on PR #67

### DIFF
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -13,14 +13,17 @@ import IW.Core.Repo (RepoName (..), RepoOwner (..))
 -- | Data type representing a GitHub issue.
 data Issue = Issue
     { issueId        :: Id Issue
+    , issueRepoOwner :: RepoOwner
+    , issueRepoName  :: RepoName
     , issueNumber    :: Int
     , issueTitle     :: Text
     , issueBody      :: Text
-    , issueRepoOwner :: RepoOwner
-    , issueRepoName  :: RepoName
     , issueLabels    :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow, ToRow)
+      deriving anyclass (ToJSON, FromRow)
+
+instance ToRow Issue where
+    toRow Issue{..} = toRow (issueRepoOwner, issueRepoName, issueNumber, issueTitle, issueBody, issueLabels)
 
 issueUrl :: Issue -> Text
 issueUrl Issue{..} = 
@@ -30,3 +33,4 @@ issueUrl Issue{..} =
     <> unRepoName issueRepoName 
     <> "/issues/" 
     <> show issueNumber
+    
--- a/src/IW/Db/Issue.hs
+++ b/src/IW/Db/Issue.hs
@@ -3,10 +3,10 @@
 -- | SQL queries to work with the @issues@ table.
 
 module IW.Db.Issue
-       ( upsertIssues
-       , getIssues
+       ( getIssues
        , getIssueById
        , getIssuesByLabel
+       , upsertIssues
        ) where
 
 import IW.App (WithError)
@@ -42,7 +42,12 @@ getIssuesByLabel label = query [sql|
 upsertIssues :: (WithDb env m) => [Issue] -> m ()
 upsertIssues issues = executeMany [sql|
     INSERT INTO issues (repo_owner, repo_name, number, title, body, labels)
-    VALUES (?, ?, ?, ?, ?, ?)
-    ON CONFLICT (repo_owner, repo_name, number) DO 
+    SELECT repo_owner, repo_name, number, title, body, labels
+    FROM (VALUES (?, ?, ?, ?, ?, ?))
+    AS new (repo_owner, repo_name, number, title, body, labels)
+    WHERE EXISTS (
+        SELECT (owner, name) FROM repos WHERE (repos.owner, repos.name) = (new.repo_owner, new.repo_name)
+    )
+    ON CONFLICT ON CONSTRAINT unique_issues DO 
     UPDATE SET title = EXCLUDED.title, body = EXCLUDED.body, labels = EXCLUDED.labels;
 |] issues
--- a/src/IW/Db/Issue.hs
+++ b/src/IW/Db/Issue.hs
@@ -3,7 +3,8 @@
 -- | SQL queries to work with the @issues@ table.
 
 module IW.Db.Issue
-       ( getIssues
+       ( insertIssues
+       , getIssues
        , getIssueById
        , getIssuesByLabel
        ) where
@@ -11,19 +12,19 @@ module IW.Db.Issue
 import IW.App (WithError)
 import IW.Core.Issue (Issue (..))
 import IW.Core.Id (Id (..))
-import IW.Db.Functions (WithDb, asSingleRow, query, queryRaw)
+import IW.Db.Functions (WithDb, asSingleRow, executeMany, query, queryRaw)
 
 -- | Returns all issues in the database
 getIssues :: (WithDb env m) => m [Issue]
 getIssues = queryRaw [sql|
-    SELECT id, number, title, body, repo_owner, repo_name, labels
+    SELECT id, repo_owner, repo_name, number, title, body, labels
     FROM issues
 |]
 
 -- | Returns a single issue with the corresponding ID
 getIssueById :: (WithDb env m, WithError m) => Id Issue -> m Issue
 getIssueById issueId = asSingleRow $ query [sql|
-    SELECT id, number, title, body, repo_owner, repo_name, labels
+    SELECT id, repo_owner, repo_name, number, title, body, labels
     FROM issues
     WHERE id = ?
 |] (Only issueId)
@@ -31,7 +32,14 @@ getIssueById issueId = asSingleRow $ query [sql|
 -- | Returns a list of issues filtered by label name
 getIssuesByLabel :: (WithDb env m) => Text -> m [Issue]
 getIssuesByLabel label = query [sql|
-    SELECT id, number, title, body, repo_owner, repo_name, labels 
+    SELECT id, repo_owner, repo_name, number, title, body, labels
     FROM issues 
     WHERE ? = ANY (labels)
 |] (Only label)
+
+-- | Insert a list of issues into the database
+insertIssues :: (WithDb env m) => [Issue] -> m ()
+insertIssues issues = executeMany [sql|
+    INSERT INTO issues (repo_owner, repo_name, number, title, body, labels)
+    VALUES (?, ?, ?, ?, ?, ?)
+|] issues
--- a/src/IW/Db/Issue.hs
+++ b/src/IW/Db/Issue.hs
@@ -4,15 +4,12 @@
 
 module IW.Db.Issue
        ( getIssues
-       , getIssueById
        , getIssuesByLabel
        , upsertIssues
        ) where
 
-import IW.App (WithError)
 import IW.Core.Issue (Issue (..))
-import IW.Core.Id (Id (..))
-import IW.Db.Functions (WithDb, asSingleRow, executeMany, query, queryRaw)
+import IW.Db.Functions (WithDb, executeMany, query, queryRaw)
 
 
 -- | Returns all issues in the database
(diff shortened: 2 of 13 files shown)


rashadg1030:
Conflicts resolved. Ready to merge.

src/IW/Core/Issue.hs
```diff
@@ -19,4 +19,8 @@ data Issue = Issue
     , issueRepoName :: Text
     , issueLabels   :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow, ToRow)
+      deriving anyclass (ToJSON, FromRow)
+
+instance ToRow Issue where
+    toRow Issue{..} = toRow (issueNumber, issueTitle, issueBody, issueUrl, issueOwner, issueRepoName, issueLabels)
```
chshersh: This instance does the work for now. But I find a couple of problems with this instance.

1. It doesn't use the `Id` type. For me, it's just strange when some fields are ignored. In my understanding, `FromRow` and `ToRow` instances should use all fields of the data type and satisfy roundtrip property-test. I had an idea for this particular problem. The solution is to introduce the following data type:
```haskell
data WithId a = WithId
    { withIdId :: Id a
    , withidVal :: a
    }
```
So, `WithId` has only `FromRow` instance, and data types like `Issue` and `Repo` doesn't store `WithId`. But we never introduced such data type in our system because it was awkward to work with.
2. You're using `RecordWildCards` here. In most cases where we use `RecordWildCards` they are perfectly fine. But here there could be a problem: somebody can introduce new fields to the `Issue` data type, and the compiler won't tell you to update this `ToRow` instance. So it could easily become broken. Again, in general case, even this is not a problem, because if you have `FromRow` instance and you have roundtrip property-based tests, you will check this. But since this instance doesn't satisfy this property and in SQL it's okay to have nullable columns so even unit tests won't catch this error, this instance is error-prone.

With `Generic` we automatically get the property of satisfying roundtrip law, and the instances will be updated automatically once you change fields. So, the solution here is to implement custom smart property-based roundtrip tests for such data types. I propose to create an issue about that, and we will discuss and implement it later.

src/IW/Db/Issue.hs
```diff
@@ -35,3 +38,17 @@ getIssuesByLabel label = query [sql|
...
+insertIssue issue = execute [sql|
+    INSERT INTO issues (number, title, body, url, owner, repo_name, labels)
+    VALUES (?, ?, ?, ?, ?, ?, ?)
+|] issue
+
+-- | Insert a list of issues into the database
+insertIssues :: (WithDb env m) => [Issue] -> m ()
+insertIssues issues = executeMany [sql|
+    INSERT INTO issues (number, title, body, url, owner, repo_name, labels)
+    VALUES (?, ?, ?, ?, ?, ?, ?)
```
chshersh: There is a potential problem with this function. You can insert issue which doesn't have the corresponding repository, so DB will throw you an error about foreign key constraint violation. This can easily happen if your async work will have the following steps:

1. Sync all repos.
2. Sync all issues.

Repositories and users can be renamed before you fetch all issues, new repos and new issues can appear before you fetched all issues but synced all repos.

Could you please try to play with SQL in REPL using INSERT with `ON CONFLICT DO NOTHING` clause to verify that if FK violates we can just ignore such issues?

* https://www.postgresql.org/docs/9.5/sql-insert.html

If it doesn't work, I propose you to play with this function in GHCi REPL to check the behavior during error: will it stop on first error or will it insert everything anyways? If we have undesired behavior, I propse to create a question on SO about this problem: How to ignore insert in case of foreign key violation?

src/IW/Core/Issue.hs
rashadg1030: Okay got it. I think I understand what we're trying to accomplish here but I'll do some more research and discuss next time we do a conference call

src/IW/Db/Issue.hs
rashadg1030: Understood. I will try to test this out and see what happens.

chshersh: Oh, one more thing I forgot. We actually need UPSERT query here. We need UPDATE when the object exists and INSERT otherwise. Check also how this is usually done in PostgreSQL.

rashadg1030: UPSERT? Okay I will research this

rashadg1030: Okay so I've done a bit of looking. It seems like there isn't a simple way to ignore the foreign key constraint and insert the rest of the issues. `ON CONFLICT DO NOTHING` still ignores the rest of the issues. I'm still using `ON CONFLICT DO UPDATE ....` to perform `UPSERT`s. This is no problem. The most common solution I can find for ignoring the foreign key constraint is first inserting the issues into something called a staging table, filtering the issues that don't have correct foreign keys, and then inserting into the actual issues table. I'm still doing more research.

rashadg1030: Apparently, you can also use triggers. Not sure which is better.  https://stackoverflow.com/questions/39827797/postgres-ignore-insert-error-and-carry-ony

rashadg1030: These also seem to be valid solutions:
https://stackoverflow.com/questions/36038751/can-insert-on-conflict-be-used-for-foreign-key-violations/36039580#36039580

https://stackoverflow.com/questions/45228853/how-to-stop-the-insert-or-update-on-table-violates-foreign-key-constraint?rq=1

rashadg1030:
Excuse me for changing the schema again, but I think this is an improvement. Going to try to implement `UPSERT` now.

rashadg1030:
We now have `UPSERT` statements for `repos` and `issues` tables.

rashadg1030:
The only thing left for this PR is deciding what to do with foreign key violations. I'll wait for a second opinion before I add any more.

rashadg1030:
Is it fine if I remove `getIssueById` and `insertIssue`?  I remember someone saying that we will only be dealing with lists of issues and not single issues, so these functions are redundant.

chshersh:
@rashadg1030 This is a bit difficult issue. Let me do some research on that and I will come back with answers to you later.

rashadg1030:
I finally found the correct insert statement we need for inserting issues:
```
INSERT INTO issues (repo_owner, repo_name, number, title, body, labels)
SELECT repo_owner, repo_name, number, title, body, labels
FROM ( VALUES ('jgm','pandoc',342,'Fix docs','Docs need to be fixed.',ARRAY['good first issue','low hanging fruit','docs'])
            , ('marry5454', 'test0', 6, 'Update config file', 'Update the server configuration file.', ARRAY['good first issue'])
            , ('john117', 'test1', 3, 'Update config file', 'Update the server configuration file.', ARRAY['good first issue', 'help wanted'])
     )
AS new (repo_owner, repo_name, number, title, body, labels)
WHERE EXISTS (
    SELECT * FROM repos WHERE (repos.owner, repos.name) = (new.repo_owner, new.repo_name)
)
ON CONFLICT (repo_owner, repo_name, number) DO 
UPDATE SET title = EXCLUDED.title, body = EXCLUDED.body, labels = EXCLUDED.labels;
```
This inserts if the foreign key exists, and will update if there is a conflict on fields that are supposed to be unique. From the test I've run it seems to work fine. From what I've read this solution is fine, but there may be issues with concurrency. For example, right after the select statement that checks for the foreign key a row is deleted, but it still tries to insert. Also an error is thrown if the same value is in the insert statement more than once. For example:
```
INSERT INTO issues (repo_owner, repo_name, number, title, body, labels)
SELECT repo_owner, repo_name, number, title, body, labels
FROM ( VALUES ('jgm','pandoc',342,'Fix docs','Docs need to be fixed.',ARRAY['good first issue','low hanging fruit','docs'])
            ,  ('jgm','pandoc',342,'Fix docs','Docs need to be fixed.',ARRAY['good first issue','low hanging fruit','docs'])
     )
AS new (repo_owner, repo_name, number, title, body, labels)
WHERE EXISTS (
    SELECT * FROM repos WHERE (repos.owner, repos.name) = (new.repo_owner, new.repo_name)
)
ON CONFLICT (repo_owner, repo_name, number) DO 
UPDATE SET title = EXCLUDED.title, body = EXCLUDED.body, labels = EXCLUDED.labels;
```

chshersh:
@rashadg1030 Good solution! This is quite nice 🙂 

> For example, right after the select statement that checks for the foreign key a row is deleted, but it still tries to insert.

Yeah, this won't happen in our case because we don't delete rows from the `repos` table

src/IW/Db/Issue.hs
```diff
@@ -14,32 +14,35 @@ import IW.Core.Issue (Issue (..))
...
 getIssues :: (WithDb env m) => m [Issue]
 getIssues = queryRaw [sql|
-    SELECT id, number, title, body, repo_owner, repo_name, labels
+    SELECT id, repo_owner, repo_name, number, title, body, labels
     FROM issues
 |]
 
 -- | Returns a single issue with the corresponding ID
 getIssueById :: (WithDb env m, WithError m) => Id Issue -> m Issue
 getIssueById issueId = asSingleRow $ query [sql|
```
chshersh: I think for now you can remove `getIssueById` function. Just to have less code to test and worry about. We can easily write it back again when we need it.

rashadg1030:
This PR should be good to go

src/IW/Db/Issue.hs
```diff
@@ -14,32 +14,35 @@ import IW.Core.Issue (Issue (..))
...
-insertIssues issues = executeMany [sql|
-    INSERT INTO issues (number, title, body, repo_owner, repo_name, url, labels)
-    VALUES (?, ?, ?, ?, ?, ?, ?)
+-- | Insert a list of issues into the database, but update on conflict
+upsertIssues :: (WithDb env m) => [Issue] -> m ()
+upsertIssues issues = executeMany [sql|
+    INSERT INTO issues (repo_owner, repo_name, number, title, body, labels)
+    VALUES (?, ?, ?, ?, ?, ?)
+    ON CONFLICT (repo_owner, repo_name, number) DO 
+    UPDATE SET title = EXCLUDED.title, body = EXCLUDED.body, labels = EXCLUDED.labels;
```
chshersh: Let's format this SQL query a bit differently to improve readability:

```sql
    INSERT INTO issues 
      (repo_owner, repo_name, number, title, body, labels)
    VALUES
      (?, ?, ?, ?, ?, ?, ?)
    ON CONFLICT (repo_owner, repo_name, number) DO 
    UPDATE SET
        title = EXCLUDED.title
      , body = EXCLUDED.body
      , labels = EXCLUDED.labels
```

rashadg1030: This is what the query looks like: 
```
    INSERT INTO issues (repo_owner, repo_name, number, title, body, labels)
    SELECT repo_owner, repo_name, number, title, body, labels
    FROM (VALUES (?, ?, ?, ?, ?, ?))
    AS new (repo_owner, repo_name, number, title, body, labels)
    WHERE EXISTS (
        SELECT (owner, name) FROM repos WHERE (repos.owner, repos.name) = (new.repo_owner, new.repo_name)
    )
    ON CONFLICT ON CONSTRAINT unique_issues DO 
    UPDATE SET 
        title = EXCLUDED.title
      , body = EXCLUDED.body
      , labels = EXCLUDED.labels;
```
Is this fine?

rashadg1030:
```
    INSERT INTO issues 
        (repo_owner, repo_name, number, title, body, labels)
    SELECT 
        repo_owner, repo_name, number, title, body, labels
    FROM (
        VALUES 
            (?, ?, ?, ?, ?, ?)
         )
    AS 
        new (repo_owner, repo_name, number, title, body, labels)
    WHERE EXISTS (
        SELECT (owner, name) FROM repos WHERE (repos.owner, repos.name) = (new.repo_owner, new.repo_name)
                 )
    ON CONFLICT ON CONSTRAINT unique_issues DO 
    UPDATE SET 
        title = EXCLUDED.title
      , body = EXCLUDED.body
      , labels = EXCLUDED.labels;
```
Or this?

chshersh: @rashadg1030 I've updated your last comment with formatting I think okay. Formatting SQL properly is a difficult task in case of complicated queries... Should be fine for now.

rashadg1030:
Made suggested changes

PR #69, opened 2019-06-28
[#58] Implement Repo type

### LINKED ISSUE #58: Implement Repo type

rashadg1030:
Resolves #58


src/IW/Core/Repo.hs
```diff
@@ -1,7 +1,19 @@
...
+import IW.Core.SqlArray (SqlArray (..))
         
 
--- | Data type representing a GitHub repository.
-data Repo
+-- | Data type representing a GitHub repository
+data Repo = Repo 
+    { repoId         :: Id Repo
+    , repoOwner      :: Text
+    , repoName       :: Text
```
chshersh: Owner and Repo name are used already in two data types. And they also used as foreign keys. So it's better at this stage to convert `Text` to newtypes. Let's introduce more newtypes inside this module:

1. `RepoOwner`
2. `RepoName`

And we can also use them in the `Issue` data type.

rashadg1030:
```
-- | Unsafe cast of 'Id'. Implementation uses smart trick to enforce usage
-- always with @TypeApplications@.
castId :: forall to from to' . ((to == to') ~ 'True) => Id from -> Id to'
castId (Id a) = Id a
```
Will `Owner` and `Name` need this cast function as well? 
Also, I will add these newtypes to the `Issue` fields in another PR after the other branches related to `Issue` are merged

chshersh: @rashadg1030 These newtypes don't have phantom type parameters. In our system `Owner` can be only for `Repo` so let's have just `newtype RepoOwner` for simplicity. Same for the `Name`. We won't have a user data type for quite a long time. so we will have only `RepoName` We can refactor data types easily later after we have more types and we can notice the pattern. But for now let's just have simple types so type system won't get in the way.

rashadg1030:
Fixed

rashadg1030:
Rebase

issue-wanted.cabal
```diff
@@ -72,6 +72,8 @@ library
                          -- Core Modules
                          IW.Core.Id
                          IW.Core.Issue
+                         IW.Core.RepoName
+                         IW.Core.RepoOwner
```
chshersh: In general, it's a good practice to create a separate module per data type. But in our case, we have just two small newtypes. You can put them inside `Repo` module directly. We can refactor them later after we start noticing some cyclic dependencies. But there is a big chance we won't have any new data types for a long type. So you will just end up writing 4 lines of imports every time you need to deal with these types and it's just not convenient.

src/IW/Core/RepoName.hs
```diff
@@ -0,0 +1,13 @@
+{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
+
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
chshersh: These two extensions and options are not needed

rashadg1030: Fixed

PR #71, opened 2019-06-28
[#62] Implement database related functions for Repo type

### LINKED ISSUE #62: Implement database related functions for Repo type

rashadg1030:
Resolves #62

Depends on PR #69

### DIFF
--- a/src/IW/Core/Repo.hs
+++ b/src/IW/Core/Repo.hs
@@ -27,7 +27,10 @@ data Repo = Repo
     , repoDescr      :: !Text
     , repoCategories :: !(SqlArray Text)
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow, ToRow)
+      deriving anyclass (ToJSON, FromRow)
+
+instance ToRow Repo where
+    toRow Repo{..} = toRow (repoOwner, repoName, repoDescr, repoCategories)
 
 repoUrl :: Repo -> Text
 repoUrl Repo{..} = 
--- a/src/IW/Core/Repo.hs
+++ b/src/IW/Core/Repo.hs
@@ -27,10 +27,7 @@ data Repo = Repo
     , repoDescr      :: !Text
     , repoCategories :: !(SqlArray Text)
     } deriving stock (Generic, Show, Eq)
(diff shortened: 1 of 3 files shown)


src/IW/Db/Repo.hs
```diff
@@ -18,6 +19,14 @@ getRepos = queryRaw [sql|
     FROM repos
 |]
 
+-- | Returns a list of repos filtered by category name
+getReposByCategory :: (WithDb env m) => Text -> m [Repo]
+getReposByCategory category = query [sql|
+    SELECT id, owner, name, descr, categories
+    FROM repos
+    WHERE ? = ANY (categories)
```
chshersh: This is okay for now. But in future, we will need to implement more complicated logic here. Because on the frontend, it should be possible to specify multiple categories and query all issues that has one of the specified categories. Same for labels. This will become relevant once we implement API endpoints and integration with the frontend. But you can create an issue about that so that we will keep track of such details.

src/IW/Db/Repo.hs
```diff
@@ -18,6 +19,14 @@ getRepos = queryRaw [sql|
     FROM repos
 |]
 
+-- | Returns a list of repos filtered by category name
+getReposByCategory :: (WithDb env m) => Text -> m [Repo]
```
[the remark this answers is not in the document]
rashadg1030: Good idea :)

rashadg1030:
Made suggested changes

src/IW/Core/Repo.hs
```diff
@@ -27,7 +27,10 @@ data Repo = Repo
     , repoDescr      :: !Text
     , repoCategories :: !(SqlArray Text)
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow, ToRow)
+      deriving anyclass (ToJSON, FromRow)
+
+instance ToRow Repo where
+    toRow Repo{..} = toRow (repoOwner, repoName, repoDescr, repoCategories)
```
chshersh: I think this instance should be derived. Probably this was an issue with resolving conflicts...

rashadg1030: Oops. Fixed it.

rashadg1030:
Fixed

PR #75, opened 2019-06-30
[#74] Setup round-trip property testing for DB

### LINKED ISSUE #33: Setup testing suite
I would like to work on setting up the testing suite before I start making any serious changes to the code. Should I refer to the `three-layer` example tests? If so, will we need `hedgehog` for testing or is `hspec` enough?

### LINKED ISSUE #74: Setup round-trip property testing for DB functions
Refers to #33

rashadg1030:
Resolves #74 
Resolves #33

### DIFF
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -121,6 +121,20 @@ test-suite issue-wanted-test
   import:              common-options
   type:                exitcode-stdio-1.0
   hs-source-dirs:      test
+
   main-is:             Spec.hs
+  other-modules:       Test.Assert
+                       Test.Common
+                       Test.Core.Issue
+                       Test.Core.Repo
+                       Test.Db
+
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
   build-depends:       issue-wanted
+                     , hedgehog ^>= 1.0
+                     , hspec
+                     , hspec-core
+                     , mtl
+                     , postgresql-simple
                      , resource-pool
+                     , unordered-containers
--- a/sql/schema.sql
+++ b/sql/schema.sql
@@ -6,25 +6,25 @@
 -----------------
 
 CREATE TABLE IF NOT EXISTS repos
-( id         SERIAL    PRIMARY KEY  
-, owner      TEXT      NOT NULL       
-, name       TEXT      NOT NULL
-, descr      TEXT      NOT NULL
-, categories TEXT      ARRAY NOT NULL
-, created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
-, updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
+( id         SERIAL      PRIMARY KEY  
+, owner      TEXT        NOT NULL       
+, name       TEXT        NOT NULL
+, descr      TEXT        NOT NULL
+, categories TEXT ARRAY  NOT NULL
+, created_at TIMESTAMP   WITH TIME ZONE DEFAULT NOW() NOT NULL
+, updated_at TIMESTAMP   WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
 
 CREATE TABLE IF NOT EXISTS issues
-( id         SERIAL    PRIMARY KEY 
-, number     INT       NOT NULL
-, title      TEXT      NOT NULL
-, body       TEXT      NOT NULL
-, repo_owner TEXT      NOT NULL
-, repo_name  TEXT      NOT NULL
-, labels     TEXT      ARRAY NOT NULL
-, created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
-, updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
+( id         SERIAL      PRIMARY KEY 
+, number     INT         NOT NULL
+, title      TEXT        NOT NULL
+, body       TEXT        NOT NULL
+, repo_owner TEXT        NOT NULL
+, repo_name  TEXT        NOT NULL
+, labels     TEXT ARRAY  NOT NULL
+, created_at TIMESTAMP   WITH TIME ZONE DEFAULT NOW() NOT NULL
+, updated_at TIMESTAMP   WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
 
 -----------------------------
--- a/sql/schema.sql
+++ b/sql/schema.sql
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS repos
 , owner      TEXT      NOT NULL       
 , name       TEXT      NOT NULL
 , descr      TEXT      NOT NULL
-, categories TEXT      ARRAY
+, categories TEXT      ARRAY NOT NULL
 , created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 , updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
@@ -22,7 +22,7 @@ CREATE TABLE IF NOT EXISTS issues
 , body       TEXT      NOT NULL
 , repo_owner TEXT      NOT NULL
 , repo_name  TEXT      NOT NULL
-, labels     TEXT      ARRAY
+, labels     TEXT      ARRAY NOT NULL
 , created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 , updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -20,7 +20,10 @@ data Issue = Issue
     , issueRepoName  :: RepoName
     , issueLabels    :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow, ToRow)
+      deriving anyclass (ToJSON, FromRow)
+
+instance ToRow Issue where
+    toRow Issue{..} = toRow (issueNumber, issueTitle, issueBody, issueRepoOwner, issueRepoName, issueLabels)
 
 issueUrl :: Issue -> Text
 issueUrl Issue{..} = 
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -1,7 +1,10 @@
 module Main where
 
 import Control.Exception (bracket)
+import Hedgehog (Group (..), checkParallel)
 import System.IO (hSetEncoding, utf8)
+import Test.Hspec (Spec, hspec)
+import Test.Hspec.Core.Spec (sequential)
 
 import IW (mkAppEnv)
 import IW.App (AppEnv, Env (..))
@@ -9,8 +12,22 @@ import IW.Config (loadConfig)
 import IW.Db (prepareDb)
 import IW.Effects.Log (runAppLogIO_)
 
+import Test.Common (joinSpecs)
+import Test.Core.Issue (issueRoundtripProp)
+
 import qualified Data.Pool as Pool
 
+
+hspecTests :: AppEnv -> Spec
+hspecTests = sequential . joinSpecs "issue-wanted" []
+
+hedgehogTests :: AppEnv -> Group
+hedgehogTests env = Group "Roundtrip properties" 
+    [ issueRoundtripProp env `named` "fromRow . toRow ≡ id"
+    ]
+  where
+    named :: a -> b -> (b, a)
+    named = flip (,)
 
 main :: IO ()
 main = bracket
@@ -26,3 +43,7 @@ main = bracket
 
         -- setup DB tables
         runAppLogIO_ env prepareDb
+
+        -- run all tests
+        hspec $ hspecTests env
+        ifM (checkParallel $ hedgehogTests env) exitSuccess exitFailure
--- a/test/Test/Assert.hs
+++ b/test/Test/Assert.hs
@@ -1,0 +1,55 @@
+{- | We use @hspec@ testing framework and it doesn't support monad transformers.
+At this moment there's no testing framework that supports monad transformers. So
+we need to pass @AppEnv@ manually to every function.
+All functions take `AppEnv` as last argument. Like this one:
+@
+satisfies :: App a -> (a -> Bool) -> AppEnv -> Expectation
+@
+Because of that, there're multiple ways to write tests:
+@
+1. satisfies action isJust env
+2. (action `satisfies` isJust) env
+3. action `satisfies` isJust $ env
+4. env & action `satisfies` isJust
+@
+We can go even further and introduce fancy operators, so this code can be
+written in more concise way. But this is TBD.
+@
+env & action @? isJust
+@
+-}
+
+module Test.Assert
+       ( failsWith
+       , satisfies
+       , succeeds
+       , equals
+       ) where
+
+import Test.Hspec (Expectation, expectationFailure, shouldBe, shouldSatisfy)
+
+import IW.App (App, AppEnv, AppError (..), AppErrorType, runAppAsIO)
+
+
+-- | Checks that given action runs successfully.
+succeeds :: (Show a) => App a -> AppEnv -> Expectation
+succeeds = (`satisfies` const True)
+
+-- | Checks whether return result of the action satisfies given predicate.
+satisfies :: (Show a) => App a -> (a -> Bool) -> AppEnv -> Expectation
+satisfies app p env = runAppAsIO env app >>= \case
+    Left e  -> expectationFailure $ "Expected 'Success' but got: " <> show e
+    Right a -> a `shouldSatisfy` p
+
+-- | Checks whether action fails and returns given error.
+failsWith :: (Show a) => App a -> AppErrorType -> AppEnv -> Expectation
+failsWith app err env = runAppAsIO env app >>= \case
+    Left AppError{..} -> appErrorType `shouldBe` err
+    Right a -> expectationFailure $
+        "Expected 'Failure' with: " <> show err <> " but got: " <> show a
+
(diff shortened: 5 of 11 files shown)


issue-wanted.cabal
```diff
@@ -121,6 +121,21 @@ test-suite issue-wanted-test
...
+                       Test.Core.Issue
+                       Test.Core.Repo
+                       Test.Db
+
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
   build-depends:       issue-wanted
+                     , hedgehog
+                     , hspec
+                     , mtl
+                     , hedgehog ^>= 0.6.1
```
chshersh: `hedgehog` is specified twice in the `build-depends` list. Also, let's use `1.0` version.

test/Test/Assert.hs
```diff
@@ -0,0 +1,59 @@
...
+    Left e  -> expectationFailure $ "Expected 'Success' but got: " <> show e
+    Right a -> a `shouldSatisfy` p
+
+-- | Checks whether action fails and returns given error.
+failsWith :: (Show a) => App a -> AppErrorType -> AppEnv -> Expectation
+failsWith app err env = runAppAsIO env app >>= \case
+    Left e@AppError{..} ->
+        if appErrorType == err
+            then appErrorType `shouldBe` err
+            else expectationFailure $
```
chshersh: This actually can be written simpler. I wrote it this way I don't remember the reason... @vrom911 noticed that here you can just write the following after `->`:

```haskell
appErrorType `shouldBe` err
```

If you have time to fix this in `three-layer` as well, it would be lovely :slightly_smiling_face:

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+    number <- genNumber
+    title <- genTitle
+    body <- genBody
+    repoOwner <- genRepoOwner
+    repoName <- genRepoName
+    labels <- genLabels
+    -- | Construct issue url
+    let url = "https://github.com/" <> repoOwner <> "/" <> repoName <> "/issues/" <> show number
+
+    pure $ Issue id' number title body url repoOwner repoName labels
```
chshersh: You can use the other side of `RecordWildCards` here. You can use it on creation if you name variables exactly like fields:

```haskell
issueId     <- genIssueId
issueNumber <- genIssueNumber
...
pure Issue{..}
```
Since all fields inside the `Issue` data type are strict, you will get an error if you forgot to initialise some fields :smirk:

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+    body <- genBody
+    repoOwner <- genRepoOwner
+    repoName <- genRepoName
+    labels <- genLabels
+    -- | Construct issue url
+    let url = "https://github.com/" <> repoOwner <> "/" <> repoName <> "/issues/" <> show number
+
+    pure $ Issue id' number title body url repoOwner repoName labels 
+    where
+        genId :: MonadGen m => m (Id Issue)
```
chshersh: The indentation here is wrong. It should be 2 spaces before `where` and 4 spaces before functions inside `where`

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+        genId = Id <$> Gen.int (Range.constant 8 40)
+
+        genNumber :: MonadGen m => m Int
+        genNumber = Gen.int (Range.constant 8 40)
+   
+        genTitle :: MonadGen m => m Text
+        genTitle = Gen.text (Range.constant 8 40) Gen.alpha
+
+        genBody :: MonadGen m => m Text 
+        genBody = Gen.text (Range.constant 8 40) Gen.alphaNum
```
chshersh: Btw, `body` in your table is a nullable column, so the field actually should have `Maybe Text` type. Alternatively, you can have `NOT NULL` body and empty body is just empty text. In that case `Range` for this generator should start from 0.

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+
+        genRepoOwner :: MonadGen m => m Text
+        genRepoOwner = Gen.text (Range.constant 8 40) Gen.alpha
```
chshersh: I propose to start it from `1` because the owner can be one character long. Also, the generator should be `alphaNum` because user names can contain digits.

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+import IW.Core.Issue (Issue (..))
+import IW.Core.SqlArray (SqlArray (..))
+
+import Hedgehog
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+testLabels :: [Text]
+testLabels = [ "good first issue"
+             , "help wanted"
```
chshersh: Let's format lists from a new line. So it will:

```haskell
testLabels =
    [ "good first issue"
    ...
    ]
```

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+        genTitle = Gen.text (Range.constant 8 40) Gen.alpha
+
+        genBody :: MonadGen m => m Text 
+        genBody = Gen.text (Range.constant 8 40) Gen.alphaNum
+
+        genRepoOwner :: MonadGen m => m Text
+        genRepoOwner = Gen.text (Range.constant 8 40) Gen.alpha
+        
+        genRepoName :: MonadGen m => m Text
+        genRepoName = Gen.text (Range.constant 8 40) Gen.alpha
```
chshersh: Some of these generators will be moved into some common `Test.Gen` module to be reused in both `Issue` and `Repo` but for now it's okay to have them here.

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+
+        genLabels :: MonadGen m => m (SqlArray Text)
+        genLabels = SqlArray <$> Gen.subsequence testLabels
```
chshersh: Strictly speaking, all labels are unique. But I think it's okay to skip this check in tests. GitHub API will return list of unique labels so we don't need to worry about this in our code.

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,52 @@
...
+genIssue = do
+    id' <- genId
+    number <- genNumber
+    title <- genTitle
+    body <- genBody
+    repoOwner <- genRepoOwner
+    repoName <- genRepoName
+    labels <- genLabels
+    -- | Construct issue url
+    let url = "https://github.com/" <> repoOwner <> "/" <> repoName <> "/issues/" <> show number
```
chshersh: I expect this function to be useful in our code as well, so you can already move it to the `IW.Core.Issue` module.

rashadg1030:
Got it. I’ll fix all these ASAP

test/Test/Assert.hs
rashadg1030: Got it!

test/Test/Core/Issue.hs
rashadg1030: This is cool. Didn't know this.

rashadg1030:
@chshersh @vrom911 So I was able to set up a test. but probably didn't do it correctly. There are a few issues with the test. Before I move any further I just wanna make sure I'm going the right direction.

chshersh:
@rashadg1030 There are several problems with your approach.

1. As you can see from the test error message, you violate foreign key constraints. You don't have any repos in your database so you can't just insert issues. Setting up a property-based testing system for the whole application is a pain, so usually, we don't do this, we use property-based testing to check laws, roundtrip properties but not the state of the whole application. There is ongoing research in this direction (google `haskell State-machine property testing`) but the approach is still immature and will require a lot of effort to implement. And our system is not that critical in terms of correctness 🙂
2. You insert from 1 to 30 issues every time, but since you're working with the same database, all issues will be accumulated there and never deleted. So querying all issues from the database will return more and more issues with every run and property will always fail.

So you actually don't have to insert issues in the database. In my humble opinion, it would be nice if the design of `postgresql-simple` allows having just pure property tests like `fromRow . toRow = id`. However, it doesn't provide such nice thing. But you can write your test like this:

```haskell
let generatedIssue = Issue{..}
Only parsedIssue <- asSingleRow $ query [sql|SELECT ?, ?, ?|] generatedIssue  -- user proper number of ? here
generatedIssue === parsedIssue
```

rashadg1030:
> @rashadg1030 There are several problems with your approach.
> 
> 1. As you can see from the test error message, you violate foreign key constraints. You don't have any repos in your database so you can't just insert issues. Setting up a property-based testing system for the whole application is a pain, so usually, we don't do this, we use property-based testing to check laws, roundtrip properties but not the state of the whole application. There is ongoing research in this direction (google `haskell State-machine property testing`) but the approach is still immature and will require a lot of effort to implement. And our system is not that critical in terms of correctness slightly_smiling_face
> 2. You insert from 1 to 30 issues every time, but since you're working with the same database, all issues will be accumulated there and never deleted. So querying all issues from the database will return more and more issues with every run and property will always fail.
> 
> So you actually don't have to insert issues in the database. In my humble opinion, it would be nice if the design of `postgresql-simple` allows having just pure property tests like `fromRow . toRow = id`. However, it doesn't provide such nice thing. But you can write your test like this:
> 
> ```haskell
> let generatedIssue = Issue{..}
> Only parsedIssue <- asSingleRow $ query [sql|SELECT ?, ?, ?|] generatedIssue  -- user proper number of ? here
> generatedIssue === parsedIssue
> ```

Regarding point 2 that you made, when the test is started I thought the `prepareDb` function runs and drops all tables then seeds them again? Either way, I will fix the test to work with a single issue instead of a list. Thank you!

rashadg1030:
```
let generatedIssue = Issue{..}
Only parsedIssue <- asSingleRow $ query [sql|SELECT ?, ?, ?|] generatedIssue  -- user proper number of ? here
generatedIssue === parsedIssue
```
Trying to replicate this code snippet right here, but no luck. The test throws this error: 
```
FormatError {fmtMessage = "8 single '?' characters, but 7 parameters", fmtQuery = "SELECT ?, ?, ?, ?, ?, ?, ?, ?", fmtParams = ["156",
"rDbuKbqIMqOOBdusNKJwAYrmsoulw","OUKsd25OrKFe8wCyGxPwJevw7Jr1DaoeV","fPtSfaLZz1X65YO9OQ","Ql2KAmDNgcMlgTqYs","https://github.com/fPtSfaLZz1X65YO9OQ
/Ql2KAmDNgcMlgTqYs/issues/156","ARRAY[help wanted,low hanging fruit,docs,easy]"]}
```

I think I need to change the `ToRow` isntance of the `Issue` type.

chshersh:
@rashadg1030 

> Regarding point 2 that you made, when the test is started I thought the prepareDb function runs and drops all tables then seeds them again?

Yes, `prepareDb` function will cleanup the DB before running all tests. But property-based tests are implemented that way that they will generate about 100 tests during run of a single test-suite. So your `issueInsertAndQuery` test will actually run 100 hundred times because it will generate an `Issue` multiple times.

src/IW/Core/Issue.hs
```diff
@@ -20,4 +20,7 @@ data Issue = Issue
     , issueUrl       :: Text
     , issueLabels    :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow, ToRow)
+      deriving anyclass (ToJSON, FromRow)
+
+instance ToRow Issue where
+  toRow Issue{..} = toRow (issueNumber, issueTitle, issueBody, issueRepoOwner, issueRepoName, issueUrl, issueLabels)
```
chshersh: Here should be 4 spaces

rashadg1030:
Oh wow haha got it

test/Spec.hs
```diff
@@ -1,17 +1,34 @@
...
+
 import qualified Data.Pool as Pool
 
 
+hspecTests :: AppEnv -> Spec
+hspecTests = sequential . joinSpecs "issue-wanted" []
+
+hedgehogTests :: AppEnv -> Group
+hedgehogTests env = Group "Roundtrip properties" 
+    [ issueInsertAndQuery env `named` "issues === (insert . query $ issues)"
```
chshersh: The text here is outdated a little bit. It's not a single issue. Also, for simplicity reasons I propose to not mention `query` here. It's an implementation detail of how we test roundtrip properties. So the text can be just:

```haskell
fromRow . toRow ≡ id
```

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,69 @@
...
+import qualified Hedgehog.Range as Range
+
+
+issueInsertAndQuery :: AppEnv -> Property
+issueInsertAndQuery env = property $ do
+  generatedIssue <- forAll genIssue
+  parsedIssue <- liftIO 
+              $ runAppLogIO env  
+              $ asSingleRow
+              $ query [sql| SELECT ?, ?, ?, ?, ?, ?, ?, ?|] generatedIssue
```
chshersh: Yeah, we almost there. Just one little tweak left. The problem now is because `fromRow` for `Issue` expects `Id` but `toRow` doesn't put `Id`. But we also don't want `toRow` to put `Id`! So the solution is to put the `Id` inside `SELECT`. Fortunately, this can be done relatively easy using data type composition from `postgresql-simple`:

```haskell
query [sql| SELECT ?, ?, ?, ?, ?, ?, ?, ?|] (Only (issueId generatedIssue) :. generatedIssue)
```

Yeah, at this point I think that our property-based roundtrip testing with `posgresql-simple` deserves a separate blog post...

rashadg1030: I agree! It's an interesting problem.

rashadg1030: I'm going to try to implement this.

rashadg1030:
Ok it seems to be passing some tests, but it eventually fails. I'm trying to figure out what's wrong.

rashadg1030:
```
 ✗ fromRow . toRow ≡ id failed after 83 tests.
  
       ┏━━ test/Test/Core/Issue.hs ━━━
    20 ┃ issueRoundtripProp :: AppEnv -> Property
    21 ┃ issueRoundtripProp env = property $ do
    22 ┃     generatedIssue <- forAll genIssue
    23 ┃     parsedIssue <- liftIO 
    24 ┃                 $ runAppLogIO env  
    25 ┃                 $ asSingleRow
    26 ┃                 $ query [sql| SELECT ?, ?, ?, ?, ?, ?, ? |] (Only (issueId generate
dIssue) :. generatedIssue)
    27 ┃     parsedIssue === Right generatedIssue
       ┃     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       ┃     │ ━━━ Exception (ResultError) ━━━
       ┃     │ Incompatible {errSQLType = "text", errSQLTableOid = Nothing, errSQLField = "?
column?", errHaskellType = "PGArray Text", errMessage = ""}
  
    This failure can be reproduced by running:
    > recheck (Size 82) (Seed 6549426097029610971 4625749683479689985) fromRow . toRow ≡ id
  
  ✗ 1 failed.
```

rashadg1030:
Got past 83 tests!

chshersh:
@rashadg1030 Try to make `Array` NOT NULL in your SQL schema. Does it fix the issue? And double-check the order of fields in both `FromRow` and `ToRow` instances.

rashadg1030:
> @rashadg1030 Try to make `Array` NOT NULL in your SQL schema. Does it fix the issue? And double-check the order of fields in both `FromRow` and `ToRow` instances.

The `ToRow` fields are in the right order, but we're deriving the `FromRow` instance. Should I implement a custom `FromRow` instance?

rashadg1030:
The test passed on CI but I can't them to pass locally. It might just be me.

chshersh:
@rashadg1030 I've played a little bit with the problem. Looks like `PGArray` itself from `postgresql-simple` doesn't satisfy roundtrip property. This property for `PGArray` fails on an empty array. I tried to generate non-empty array always and test always pass. I tried to return empty array always and the tests fail on the first attempt every time.

I propose you to open an issue to the `postgresql-simple` library:

* https://github.com/phadej/postgresql-simple/issues/new

The fact that CI passed is just luck :) Looks like this time 100 times non-empty array was generated. But I can make tests fail locally consistently.

rashadg1030:
Ahh okay nice. I will open an issue on postgresql-simple.

sql/schema.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS repos
 , owner      TEXT      NOT NULL       
 , name       TEXT      NOT NULL
 , descr      TEXT      NOT NULL
-, categories TEXT      ARRAY
+, categories TEXT      ARRAY NOT NULL
```
chshersh: I think that `ARRAY` should be closer to type rather than to constraint. So shit should be something like:

```sql
categories TEXT ARRAY                 NOT NULL
```

test/Test/Core/Issue.hs
```diff
@@ -0,0 +1,69 @@
...
+import qualified Hedgehog.Range as Range
+
+
+issueRoundtripProp :: AppEnv -> Property
+issueRoundtripProp env = property $ do
+    generatedIssue <- forAll genIssue
+    parsedIssue <- liftIO 
+                $ runAppLogIO env  
+                $ asSingleRow
+                $ query [sql| SELECT ?, ?, ?, ?, ?, ?, ? |] (Only (issueId generatedIssue) :. generatedIssue)
```
chshersh: This line is really long. I propose to move it into separate function like:

```haskell
issueViaSql :: WithDb m => Issue -> m Issue
issueViaSql issue = asSingleRow $ query
    [sql| SELECT ?, ?, ?, ?, ?, ?, ? |]
    (Only (issueId generatedIssue) :. generatedIssue)
```

After that it should be possible to write this in a single line:

```haskell
    parsedIssue <- liftIO $ runAppLogIO env $ issueViaSql issue
```

rashadg1030:
Fix

PR #79, opened 2019-07-01
[#78] Update Issue SQL schema

### LINKED ISSUE #78: Remove URL field from Issue schema
I noticed something important:
```
CREATE TABLE IF NOT EXISTS repos
( id         SERIAL PRIMARY KEY  
, owner      TEXT   NOT NULL       
, name       TEXT   NOT NULL
, descr      TEXT   NOT NULL
, categories TEXT   ARRAY
);

CREATE TABLE IF NOT EXISTS issues
( id         SERIAL PRIMARY KEY 
, number     INT    NOT NULL
, title      TEXT   NOT NULL
, body       TEXT   NOT NULL
, repo_owner TEXT   NOT NULL
, repo_name  TEXT   NOT NULL
, url        TEXT   NOT NULL
, labels     TEXT   ARRAY
);
```
For `repos` table we don't have an `url` field because we agreed we could construct the URL on the frontend with the `owner` and `name`.

The `issues` table does have an `url` field but we could remove it like `repos`.

```
-- | Data type representing a GitHub issue.
data Issue = Issue
    { issueId        :: Id Issue
    , issueNumber    :: Int
    , issueTitle     :: Text
    , issueBody      :: Text
    , issueRepoOwner :: RepoOwner
    , issueRepoName  :: RepoName
    , issueUrl       :: Text
    , issueLabels    :: SqlArray Text
      deriving anyclass (ToJSON, FromRow, ToRow)
```

```
-- | Data type representing a GitHub repository
data Repo = Repo 
    { repoId         :: Id Repo
    , repoOwner      :: RepoOwner
    , repoName       :: RepoName
    , repoDescr      :: Text
    , repoCategories :: SqlArray Text
      deriving anyclass (ToJSON, FromRow, ToRow)
```
Notice the `Repo` type doesn't have `url` either. Which one is better? Construct the URL on the frontend, backend, or store it in the database?

rashadg1030:
Resolves #78

### DIFF
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -2,6 +2,7 @@
 
 module IW.Core.Issue
        ( Issue (..)
+       , issueUrl
        ) where
   
 import IW.Core.Id (Id (..))
@@ -20,3 +21,11 @@ data Issue = Issue
     , issueLabels    :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
       deriving anyclass (ToJSON, FromRow, ToRow)
+
+issueUrl :: Issue -> Text
+issueUrl Issue{..} = "https://github.com/" 
+                  <> unRepoOwner issueRepoOwner 
+                  <> "/" 
+                  <> unRepoName issueRepoName 
+                  <> "/issues/" 
+                  <> show issueNumber
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -17,7 +17,6 @@ data Issue = Issue
     , issueBody      :: Text
     , issueRepoOwner :: RepoOwner
     , issueRepoName  :: RepoName
-    , issueUrl       :: Text
     , issueLabels    :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
       deriving anyclass (ToJSON, FromRow, ToRow)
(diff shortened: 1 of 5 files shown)


rashadg1030:
Removed `url` from `Issue` type.

src/IW/Core/Issue.hs
```diff
@@ -17,7 +18,14 @@ data Issue = Issue
...
     , issueRepoOwner :: RepoOwner
     , issueRepoName  :: RepoName
-    , issueUrl       :: Text
     , issueLabels    :: SqlArray Text
     } deriving stock (Generic, Show, Eq)
       deriving anyclass (ToJSON, FromRow, ToRow)
+
+issueUrl :: Issue -> Text
+issueUrl Issue{..} = "https://github.com/" 
+                  <> unRepoOwner issueRepoOwner
```
chshersh: Let's align this pure function a little bit differently:

```haskell
issueUrl Issue{..} =
    "https://github.com/"
    <> unRepoOwner issueRepoOwner
    <> ...
```

In general we don't want to have big indentation of the following lines. So, for example, in this case to look code nicer you need to start from the line after `=`. But, for example, if you pattern-matching immediately, you can continue with the line with `=`

```haskell
issueUrl Issue{..} = case smth of  -- this is okay
    ....
    ....
```

rashadg1030:
Fixed after review.

Issue #73, opened 2019-06-29
Add created_at and updated_at columns to repos and issues tables


chshersh:
These columns should exist only inside the database. They should set to `NOW()` during creation. And `updated_at` should be updated automatically when we update the row. These columns will be useful later when we want to perform cleanup of our database.

rashadg1030:
https://www.revsys.com/tidbits/automatically-updating-a-timestamp-column-in-postgresql/

Using the information in this link as a guide.

PR #80, opened 2019-07-01
[#73] Add created_at and updated_at columns

rashadg1030:
I believe this is what we want. Not sure if we should use `WITHOUT TIME ZONE` or `WITH TIME ZONE`. Went with the latter just to be safe.

### DIFF
--- a/sql/schema.sql
+++ b/sql/schema.sql
@@ -9,7 +9,7 @@ CREATE TABLE IF NOT EXISTS repos
 ( id         SERIAL PRIMARY KEY  
 , owner      TEXT   NOT NULL       
 , name       TEXT   NOT NULL
-, descr      TEXT
+, descr      TEXT   NOT NULL
 , categories TEXT   ARRAY
 );
 
@@ -17,7 +17,7 @@ CREATE TABLE IF NOT EXISTS issues
 ( id         SERIAL PRIMARY KEY 
 , number     INT    NOT NULL
 , title      TEXT   NOT NULL
-, body       TEXT
+, body       TEXT   NOT NULL
 , repo_owner TEXT   NOT NULL
 , repo_name  TEXT   NOT NULL
 , url        TEXT   NOT NULL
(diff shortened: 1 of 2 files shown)


sql/schema.sql
```diff
@@ -6,22 +6,26 @@
...
-, owner      TEXT   NOT NULL       
-, name       TEXT   NOT NULL
-, descr      TEXT
-, categories TEXT   ARRAY
+( id         SERIAL    PRIMARY KEY  
+, owner      TEXT      NOT NULL       
+, name       TEXT      NOT NULL
+, descr      TEXT      NOT NULL
+, categories TEXT      ARRAY
+, created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW()
```
chshersh: Yeah, in all cases using `TIMESTAMP WITH TIME ZONE` is a good default. But let's alos add not-nullable constraint to every timestamp column.
```suggestion
, created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
```

sql/schema.sql
```diff
@@ -34,3 +38,18 @@ ALTER TABLE ONLY repos
 ALTER TABLE ONLY issues
   ADD CONSTRAINT fk_repos FOREIGN KEY (repo_owner, repo_name) 
   REFERENCES repos (owner, name) ON DELETE CASCADE;
+
+--------------
+-- TRIGGERS --
+--------------
+
+CREATE OR REPLACE FUNCTION update_timestamp()
```
chshersh: Yep, this is the correct way to update `updated_at` columns in PostgreSQL automatically: we need to create a trigger (custom SQL function) and call it for each row before the update. Just a minor stylistic suggestion: I would love to see `updated_at` column name in the name of this trigger, so it will be clear which one of the timestamps is updated. Because later we might have more timestamp times like when the issue was created on GitHub, what was the timestamp of the last comment, etc. So the trigger can have a name like `update_update_at` or something like that.

rashadg1030:
Fixed

Issue #86, opened 2019-07-04
Add roundtrip property test for Repo type


rashadg1030:
We need a roundtrip property test to ensure the `ToRow` and `FromRow` instances of `Repo` are correct.

PR #87, opened 2019-07-04
[#86] Add roundtrip property test for Repo

rashadg1030:


test/Spec.hs
```diff
@@ -24,6 +25,7 @@ hspecTests = sequential . joinSpecs "issue-wanted" []
 hedgehogTests :: AppEnv -> Group
 hedgehogTests env = Group "Roundtrip properties" 
     [ issueRoundtripProp env `named` "fromRow . toRow ≡ id"
+    , repoRoundtripProp env `named` "fromRow . toRow ≡ id"
```
chshersh: In terminal you will this text. So let's add some context to it to understand which test this is for:

```haskell
"Issue: fromRow . toRow ≡ id"
"Repo: fromRow . toRow ≡ id"
```

test/Test/Core/Repo.hs
```diff
@@ -1 +1,65 @@
...
+    pure Repo{..} 
+  where
+    genId :: MonadGen m => m (Id Repo)
+    genId = Id <$> Gen.int (Range.constant 1 500)
+
+    genOwner :: MonadGen m => m RepoOwner
+    genOwner = RepoOwner <$> Gen.text (Range.constant 1 20) Gen.alphaNum
+    
+    genName :: MonadGen m => m RepoName
+    genName = RepoName <$> Gen.text (Range.constant 1 20) Gen.alphaNum
```
chshersh: Let's create `Test.Gen` module and move all common generators for `Repo` and `Issue` there, so we can just import them and reuse.

rashadg1030: Is there some way I can factor out `genId` too?

chshersh: You can just generate `Id a`. For tests it doesn't matter, what type parameter is there

rashadg1030:
Made suggested changes

PR #90, opened 2019-07-05
[#89] Implement WithId type

### LINKED ISSUE #89: Implement WithId type
How will this work?

rashadg1030:
Resolves #89

My first attempt at implementing the `WithId` type. I'm still confused as to how to use `(:.)` to derive the `ToRow` and `FromRow` instances of `WithId`. I used `(<$>)` and `(<*>):
```
instance FromRow a => FromRow (WithId a) where
    fromRow = WithId <$> field <*> fromRow

instance ToRow a => ToRow (WithId a) where
    toRow WithId{..} = toRow withIdVal  
```

Another thing is the test with `WithId Issue`:

```
issueViaSql :: (WithDb env m, WithError m) => WithId Issue -> m (WithId Issue)
issueViaSql WithId{..} = asSingleRow $ query
    [sql| SELECT ?, ?, ?, ?, ?, ?, (? :: TEXT ARRAY) |]
    (Only (issueId issue) :. issue)
    ((Only withIdId) :. withIdVal)
```

I'm still using `(:.)` which is not what I want, but if I don't I get a format error saying I don't have enough parameters.

### DIFF
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -5,25 +5,20 @@ module IW.Core.Issue
        , issueUrl
        ) where
   
-import IW.Core.Id (Id (..))
 import IW.Core.SqlArray (SqlArray (..))
 import IW.Core.Repo (RepoName (..), RepoOwner (..))
 
 
 -- | Data type representing a GitHub issue.
 data Issue = Issue
-    { issueId        :: Id Issue
-    , issueRepoOwner :: RepoOwner
-    , issueRepoName  :: RepoName
-    , issueNumber    :: Int
-    , issueTitle     :: Text
-    , issueBody      :: Text
-    , issueLabels    :: SqlArray Text
+    { issueRepoOwner :: !RepoOwner
+    , issueRepoName  :: !RepoName
+    , issueNumber    :: !Int
+    , issueTitle     :: !Text
+    , issueBody      :: !Text
+    , issueLabels    :: !(SqlArray Text)
     } deriving stock (Generic, Show, Eq)
-      deriving anyclass (ToJSON, FromRow)
-
-instance ToRow Issue where
-    toRow Issue{..} = toRow (issueRepoOwner, issueRepoName, issueNumber, issueTitle, issueBody, issueLabels)
+      deriving anyclass (ToJSON, FromRow, ToRow)
 
 issueUrl :: Issue -> Text
 issueUrl Issue{..} = 
--- a/src/IW/Core/WithId.hs
+++ b/src/IW/Core/WithId.hs
@@ -1,0 +1,32 @@
+{-# LANGUAGE DeriveAnyClass #-}
+
+{- | @WithId a@ provides an @Id a@ field to types that don't have one. This was added
+because types with an @Id a@ can be sorted and updated more effeciently.
+-}
+
+module IW.Core.WithId 
+       ( WithId (..)
+       ) where
+
+import IW.Core.Id (Id (..))
+
+import Database.PostgreSQL.Simple.FromRow (RowParser)
+import Database.PostgreSQL.Simple.ToField (Action)
+import Database.PostgreSQL.Simple.Types ((:.) (..))
+
+
+data WithId a = WithId
+    { withIdId  :: !(Id a)
+    , withIdVal :: !a
+    } deriving stock (Generic, Show, Eq)
+      deriving anyclass (ToJSON)
+    
+instance FromRow a => FromRow (WithId a) where
+    fromRow :: RowParser (WithId a)
+    fromRow = WithId <$> field <*> fromRow
+    {-# INLINE fromRow #-}
+
+instance ToRow a => ToRow (WithId a) where
+    toRow :: WithId a -> [Action] 
+    toRow WithId{..} = toRow (Only withIdId :. withIdVal)  
+    {-# INLINE toRow #-}
--- a/test/Test/Core/Issue.hs
+++ b/test/Test/Core/Issue.hs
@@ -4,7 +4,7 @@ module Test.Core.Issue
        ( issueRoundtripProp
        ) where
 
-import Hedgehog
+import Hedgehog (MonadGen, Property, forAll, property, (===))
 
 import IW.App (AppEnv, WithError)
 import IW.Core.Issue (Issue (..))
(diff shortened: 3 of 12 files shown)


src/IW/Core/WithId.hs
```diff
@@ -0,0 +1,20 @@
...
+
+
+data WithId a = WithId
+    { withIdId  :: !(Id a)
+    , withIdVal :: !a
+    } deriving stock (Generic, Show, Eq)
+      deriving anyclass (ToJSON)
+    
+instance FromRow a => FromRow (WithId a) where
+    fromRow = WithId <$> field <*> fromRow
```
chshersh: Let's add `{-# INLINE #-}` pragmas to `fromRow` and `toRow` here.

src/IW/Core/WithId.hs
```diff
@@ -0,0 +1,20 @@
...
+
+instance ToRow a => ToRow (WithId a) where
+    toRow WithId{..} = toRow withIdVal
```
chshersh: Here it should be:

```suggestion
    toRow WithId{..} = toRow (Only withIdId :. withIdVal)  
```

So `WithId` actually outputs `Id` to row inside SQL query. If you don't need to convert `Id`, you just extract `withIdVal` and call `toRow` for this data type.

src/IW/Core/WithId.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE DeriveAnyClass #-}
+
+module IW.Core.WithId 
+       ( WithId (..)
+       ) where
+
+import IW.Core.Id (Id (..))
+
+
+data WithId a = WithId
```
chshersh: Let's add some documentation on this data type is needed 🙂 
We will need to write some Haddocks at some point, so other people can understand our codebase. But this can be done lazily when we are working on some features.

test/Test/Core/Issue.hs
```diff
@@ -20,10 +21,10 @@ import qualified Hedgehog.Range as Range
 import Test.Gen (genId, genRepoOwner, genRepoName)
 
 
-issueViaSql :: (WithDb env m, WithError m) => Issue -> m Issue
-issueViaSql issue = asSingleRow $ query
+issueViaSql :: (WithDb env m, WithError m) => WithId Issue -> m (WithId Issue)
+issueViaSql WithId{..} = asSingleRow $ query
     [sql| SELECT ?, ?, ?, ?, ?, ?, (? :: TEXT ARRAY) |]
-    (Only (issueId issue) :. issue)
+    ((Only withIdId) :. withIdVal)
```
chshersh: After you fix the `ToRow` instance for `WithId`, it should be possible to write this function as:

```haskel
issueViaSql = asSingleRow . query [sql| ... |]
```

src/IW/Core/Issue.hs
```diff
@@ -5,25 +5,20 @@ module IW.Core.Issue
...
 
 
 -- | Data type representing a GitHub issue.
 data Issue = Issue
-    { issueId        :: Id Issue
-    , issueRepoOwner :: RepoOwner
+    { issueRepoOwner :: RepoOwner
     , issueRepoName  :: RepoName
     , issueNumber    :: Int
     , issueTitle     :: Text
```
chshersh: While we are on this: let's make every field in our `Repo` and `Issue` data types strict.

src/IW/Core/WithId.hs
rashadg1030: Yeah right that’s what we were trying avoid initially. Makes sense👍🏽

rashadg1030:
Definitely will write some documentation for `WithId`

rashadg1030:
@chshersh I'm gonna add it to this PR right now.

src/IW/Core/WithId.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE DeriveAnyClass #-}
+
+{- | @WithId a@ wraps values of type @a@ that don't have an @Id a@ field, but need one in
+order to satisfy roundtrip property tests. We use it to avoid adding an extra field of
```
chshersh: I propose to rephrase the part about why it needs. Because the primary goal of `Withid` is to not to implement property tests but to be able to provide `Id` for a data types that don't know about their `Id`. Data types don't know about `Id` not because we wanted to implement `toRow` instance without outputting `Id` but because we fetch data from external sources and those sources don't have information about our `Id`. But later down the road, we might want to expose those `id` to the frontend, so we can do some things more efficiently, like _force repo to sync by Id_ or _update issue details by Id_.

rashadg1030: Just updated the docs. It's more concise. Let me know if it needs more information.

rashadg1030:
Just rebased on to master. Now I'm going to get adjust the functions to past the test with `WithId`.

rashadg1030:
```  
  it "should update the repo if the same repo already exists" $
        env & reposRowDifference (upsertRepos [updatedValidRepo]) `equals` zipWith tagWithId [1..] [updatedValidRepo]
```
Not sure about using `zipWith` here so that this function can be used with list of updated `Repo` values.

I could've just done 

```
  it "should update the repo if the same repo already exists" $
        env & reposRowDifference (upsertRepos [updatedValidRepo]) `equals`  [WithId (Id 1) updatedValidRepo]
```
But this is good if we only test one `updatedValidRepo`. The above solution will allow us to test multiple values, but there may be some problems with it. Let me know what you think. If we don't plan on upserting more than `Issue`/`Repo` at a time, then we don't need this solution.

test/Test/Db.hs
```diff
@@ -95,7 +100,7 @@ upsertIssuesSpec env = describe "upsertIssues" $ do
     it "should leave the issues table unaffected when there's no corresponding repo" $
         env & issuesUnaffected (upsertIssues [invalidIssue]) `equals` True 
     it "should update the issue if the same issue already exists" $
-        env & issuesRowDifference (upsertIssues [updatedValidIssue]) `equals` [updatedValidIssue]
+        env & issuesRowDifference (upsertIssues [updatedValidIssue]) `equals` zipWith tagWithId [1..] [updatedValidIssue]
```
chshersh: I think that `tagWithId` function is useful But currently our tests are pretty small. And using `zipWith` here looks like an overcomplication. Having the following looks simpler to me. Don't forget, we still need to understand code written in tets :slightly_smiling_face:  

```haskell
[Id 1 `Withid` updatedValidIssue]
```

test/Test/Gen.hs
```diff
@@ -1,9 +1,10 @@
 module Test.Gen where
 
+import Hedgehog
```
chshersh: All imports should have explicit import lists

rashadg1030: Yes, I need to fix this for some other modules as well

PR #91, opened 2019-07-06
Try to speedup CI a little bit

chshersh:
I noticed that modules in `issue-wanted` are built twice on CI. This is an attempt to improve the situtation.


rashadg1030:
Faster CI sounds good to me!

PR #95, opened 2019-07-07
[#83] Implement wrapper functions around GitHub search queries

### LINKED ISSUE #83: Add folder that holds all async worker related modules
Start setting up the file structure for the async worker code once we've figured out the database.

### LINKED ISSUE #84: Add functions that convert github libary types to our own
We will need to add functions that convert the `github` library types like `Issue` and `Repo` to our own types.

### LINKED ISSUE #96: Implement types for github library errors
The `github` library has an `Error` type that may be returned instead of a `SearchResult`. 
https://hackage.haskell.org/package/github-0.22/docs/GitHub-Data-Definitions.html#t:Error
Should we have a way of handling these errors in the App monad?

rashadg1030:
Resolves #83
Resolves #84 
Resolves #96


rashadg1030:
I decided to import `Issue`, `Repo`, and `Error` as `qualified` so we can better distinguish them from our own types. Let me know if this unnecessary.

rashadg1030:
I added a prototype function `syncAllHaskellRepos` and wanted a second opinion: 
```
-- | Fetch all repositories with Haskell language
fetchAllHaskellRepos :: IO (Either GitHub.Error (SearchResult GitHub.Repo))
fetchAllHaskellRepos = searchRepos "language:haskell"

githubErrToAppErr :: GitHub.Error -> AppErrorType
githubErrToAppErr = undefined

syncAllHaskellRepos :: (MonadIO m, WithDb env m, WithError m) => m ()
syncAllHaskellRepos = do
    result <- liftIO $ fetchAllHaskellRepos
    case result of
        Right (SearchResult count vec) -> undefined 
        Left err                       -> throwError . githubErrToAppErr $ err
```

rashadg1030:
And do we need functions like `fetchHelpWanted` or `fetchHaskellReposGFI`? All sorting and filtering will be done by the database or frontend.

src/IW/Sync/Search.hs
```diff
@@ -1,31 +1,60 @@
...
+import IW.App (WithError)
+import IW.Db (WithDb)
+import IW.App.Error (AppErrorType, throwError)
+
+import qualified GitHub (Repo, Issue, Error (..))
+
+
 -- | Fetch all repositories with Haskell language
-fetchAllHaskellRepos :: IO (Either Error (SearchResult Repo))
+fetchAllHaskellRepos :: IO (Either GitHub.Error (SearchResult GitHub.Repo))
```
chshersh: Our ultimate goal is to have function:

```haskell
fetchAllHaskellRepos :: ... m => m [Repo]
```

So you can keep existing functions and then implement wrappers around them that work nicer with our interface.

src/IW/Sync/Search.hs
```diff
@@ -1,31 +1,60 @@
...
+syncAllHaskellRepos :: (MonadIO m, WithDb env m, WithError m) => m ()
+syncAllHaskellRepos = do
+    result <- liftIO fetchAllHaskellRepos
+    case result of
+        Right (SearchResult count vec) -> undefined 
+        Left err                       -> throwError . githubErrToAppErr $ err
+        
+{-
 -- | Fetch all repositories with Haskell language and label "help wanted"
 fetchHaskellReposHW :: IO (Either Error (SearchResult Repo))
```
chshersh: Yeah, you can delete these functions, no need to have these two specializations. We have a general function to fetch issues by label.

src/IW/Sync/Search.hs
```diff
@@ -1,31 +1,60 @@
...
 fetchHaskellReposHW = searchRepos "language:haskell help-wanted-issues:>0"
 
 -- | Fetch all repositories with Haskell language and label "good first issue"
 fetchHaskellReposGFI :: IO (Either Error (SearchResult Repo))
 fetchHaskellReposGFI = searchRepos "language:haskell good-first-issues:>0"
+-}
+
+-- | Fetch all open issues with Haskell language
+fetchAllHaskellIssues :: IO (Either GitHub.Error (SearchResult GitHub.Issue))
+fetchAllHaskellIssues = searchIssues "language:haskell is:open"
```
chshersh: Could you please create an issue (for future) to implement garbage collecting of old issues? This will require to fetch all closed issues and delete all closed issues from our database. We don't need to implement this now, but it's good to keep track of that issue for future.

src/IW/Sync/Search.hs
```diff
@@ -1,31 +1,46 @@
...
+       ( fetchAllHaskellRepos
+       , fetchAllHaskellIssues
+       , fetchOpenIssuesWithLabels
+       ) where
 
-import GitHub (Error, Repo, SearchResult, Issue)
+import GitHub (SearchResult)        
 import GitHub.Endpoints.Search (searchRepos, searchIssues)
 
+import qualified GitHub (Repo, Issue, Error)
```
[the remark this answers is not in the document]
rashadg1030: Got it

src/IW/Sync/Search.hs
```diff
@@ -1,31 +1,60 @@
...
 
-import GitHub (Error, Repo, SearchResult, Issue)
+import GitHub (SearchResult (..))        
 import GitHub.Endpoints.Search (searchRepos, searchIssues)
 
+import IW.App (WithError)
+import IW.Db (WithDb)
+import IW.App.Error (AppErrorType, throwError)
+
+import qualified GitHub (Repo, Issue, Error (..))
```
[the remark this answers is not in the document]
rashadg1030: Do you mean for `IW.Db` and `IW.App.Error` too?

src/IW/Sync/Search.hs
rashadg1030: Nice

src/IW/Sync/Search.hs
rashadg1030: Good idea

src/IW/Sync/Search.hs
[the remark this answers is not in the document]
rashadg1030: Just the `qualified` got it

rashadg1030:
@chshersh 
```
githubRepoToRepo :: GitHub.Repo -> Repo
githubRepoToRepo githubRepo = Repo{..}
    where  
      repoOwner = RepoOwner . GitHub.untagName . GitHub.simpleOwnerLogin . GitHub.repoOwner $ githubRepo
      repoName  = RepoName  . GitHub.untagName . GitHub.repoName $ githubRepo
      repoDesc  = ???
```
Here I'm wondering what I should do with the `repoDescription` because in the `GitHub.Repo` type, the description is `Maybe Text` while for us it's just `Text`. In the db it's `NOT NULL`. We agreed that it would be the `""` empty string and not `Nothing`/`NULL`. So in this case should I have a function from `Maybe Text -> Text`? or introduce `Maybe` into our type?

rashadg1030:
```
githubRepoToRepo :: GitHub.Repo -> Repo
githubRepoToRepo githubRepo = Repo{..}
    where  
      repoOwner      = RepoOwner . GitHub.untagName . GitHub.simpleOwnerLogin . GitHub.repoOwner $ githubRepo
      repoName       = RepoName  . GitHub.untagName . GitHub.repoName $ githubRepo
      repoDesc       = undefined
      repoCategories = undefined -- some kind of IO is needed here
```
Another question about `repoCategories`. Because we need to download the .cabal file and parse it for the `Repo`'s categories, we either need to introduce `IO` here or just default to an empty list for now and add the categories to the database seperately.

chshersh:
> Here I'm wondering what I should do with the repoDescription
> So in this case should I have a function from Maybe Text -> Text?

Yes, in case of `Nothing` you can just assign empty text as a description.

> either need to introduce IO here or just default to an empty list for now and add the categories to the database seperately.

I propose to specify categories initially as an empty list. And we will populate them later.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,69 @@
...
+import IW.App.Error (AppErrorType, throwError)
+
+import qualified GitHub
+import qualified Data.Vector as Vector
+
+
+-- | Fetch all repositories with Haskell language
+fetchAllHaskellRepos :: (MonadIO m, WithDb env m, WithError m) => m [Repo]
+fetchAllHaskellRepos = do
+    result <- liftIO $ searchRepos "language:haskell"
```
chshersh: Code is a bit ad-hoc in this module. Let's make it cleaner. I propose to move some common and relevant parts into separate functions, so types can help us.

First, let's create function:

```haskell
searchHaskellRepos :: IO ...
searchHaskellRepos = searchRepos ...
```

I think it's useful to see the explicit type of this function and what it does.

Then, our search GitHub functions always will return some `Either ... SearchResult`. Let's introduce helper function:

```haskell
liftGitHubSearchToApp :: ... m => IO (Either ... (SearchResult a)) -> m [a]
```

This function will convert `Vector` to list and `throwError` when we see an error. This function also can do some useful logging since `SearchResult` contains total count, so we can log something like:

```
Fetching total of 100500 Repos...
```

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,69 @@
...
+-- | Fetch all open issues with Haskell language and the labels passed in to the function
+fetchOpenIssuesByLabels :: (MonadIO m, WithDb env m, WithError m) => [Label] -> m [Issue]
+fetchOpenIssuesByLabels labels = do
+    result <- liftIO $ searchIssues $ "language:haskell is:open " <> labelQualifiers
+    case result of
+        Right (SearchResult _ vec) -> return $ Vector.toList (githubIssueToIssue <$> vec)
+        Left err                   -> throwError . githubErrToAppErr $ err
+  where
+    labelQualifiers :: Text
+    labelQualifiers = foldMap (\Label{..} -> "label:\"" <> unLabel <> "\" ") labels
```
chshersh: Let's move this into a separate pure function:

```haskell
labelsToSearchQuery :: [Label] -> Text
```

rashadg1030:
I forgot to ask one more thing:
```
githubIssueToIssue :: GitHub.Issue -> Issue
githubIssueToIssue githubIssue = Issue{..}
  where  
    issueRepoOwner = undefined
    issueRepoName  = undefined
    issueNumber    = undefined 
    issueTitle     = undefined 
    issueBody      = undefined 
    issueLabels    = undefined
```
I need to implement this function, but the `GitHub.Issue` type doesn't provide `repoOwner` or `repoName`.  It has everything else though. One thing I can do to get the `repoOwner` and `repoName` is by parsing the `issueUrl` field of `GitHub.Issue`. The issue URLs have the format `https://api.github.com/repos/<repoOwner>/<repoName>/issues/<number>`

More about `GitHub.Issue` can be found here:
https://hackage.haskell.org/package/github-0.22/docs/GitHub-Data-Issues.html

chshersh:
> but the GitHub.Issue type doesn't provide repoOwner or repoName

Heh. that's awkward. In that case, indeed we need to parse `issueUrl`.

rashadg1030:
@chshersh Here's something else that's kind of odd. I'm trying to extract the `issueNumber` from `GitHub.Issue`, but it's wrapped in a newtype `IssueNumber`. The problem is that the library doesn't seem to export it's constructors, so I can't pattern match and extract the value. On hackage, there is the function `unIssueNumber` but when I try to import it I get an error saying `It's not exported by....`. Not sure what I'm doing wrong. Here's the link to the docs: https://hackage.haskell.org/package/github-0.22/docs/src/GitHub.Data.Definitions.html#IssueNumber
I tried to import it from `GitHub`, `GitHub.Data`, and `GitHub.Data.Definitions`, but no luck. I'm missing something.

chshersh:
@rashadg1030 You are reading docs for `github-0.22`, however we are using `github-0.20`:

* https://hackage.haskell.org/package/github-0.20/docs/GitHub-Data-Issues.html

In this version, there is no `IssueNumber`.

`github-0.21` was broken, but I heard that `github-0.22` fixed issues (though I never tested it).

rashadg1030:
@chshersh Okay my mistake. So we're sticking with `github-0.20` right?

chshersh:
@rashadg1030 We can use the old version for now. Or you can try to switch to `github-0.22` to not deal with migration later. I personally would love to know, whether `github-0.22` actually works, so if you have time, you can try to switch to a newer version 🙂

rashadg1030:
@chshersh Yes, I think it's a good idea to do now. I will check out the latest version and see what changes I need to make.

rashadg1030:
@chshersh Just upgrade. Didn’t need to make any changes to the code. A couple of dependencies have been added though

rashadg1030:
I implemented a "parser" for getting `repoOwner` and `repoName` from the `GitHub.Issue` but it is dirty.

```
-- | Extract the @RepoOwner@ from a value of type @GitHub.Issue@.
parseIssueUrlRepoOwner :: GitHub.Issue -> Maybe RepoOwner
parseIssueUrlRepoOwner issue = RepoOwner <$> (listToMaybe . take 1 . drop 4 . split ('/'==) . GitHub.getUrl . GitHub.issueUrl) issue 

-- | Extract the @RepoName@ from a value of type @GitHub.Issue@.
parseIssueUrlRepoName :: GitHub.Issue -> Maybe RepoName
parseIssueUrlRepoName issue = RepoName <$> (listToMaybe . take 1 . drop 5 . split ('/'==) . GitHub.getUrl . GitHub.issueUrl) issue
```
I have to introduce `Maybe`. I know we don't wanna use unsafe `head` or anything like that. How should we handle this? Everything else on this PR seems to be fine. I added some docs, but let me know if they are hard to understand. This PR is still kinda dirty, some of the lines are a bit long. I will clean it up tomorrow morning, just let me know what you would like me to add. After I clean it up I would like to close this PR. It's already pretty big and it resolves 3 issues.

chshersh:
@rashadg1030 It's okay to return `Maybe`. You can function with type `GitHub.Issue -> Maybe Issue` and then use `mapMaybe` to convert to `Issue`. Later more robust logging and error handling can be done, so we can see which issues failed to parse and why. But should be good for now. Just don't forget to create issues about possible improvements :slightly_smiling_face: So we can be aware of our tech debt and pay it later.

src/IW/App/Error.hs
```diff
@@ -120,6 +144,11 @@ toHttpError (AppError _callStack errorType) = case errorType of
         MissingHeader name     -> err401 { errBody = toLazy $ "Header not found: " <> foldedCase name }
         HeaderDecodeError name -> err401 { errBody = encodeUtf8 $ "Unable to decode header: " <> name }
         DbError e              -> err500 { errBody = encodeUtf8 e }
+    GitHubError err -> case err of
```
chshersh: I think here instead of 4 cases you can just do `show` and return `err500`. Should be okay for now.

src/IW/App/Error.hs
```diff
@@ -109,6 +112,27 @@ data IError
...
+    | JsonError Text	
+    {- | Incorrect input was provided. -}
+    | UserError Text
+    deriving (Show, Eq)
+
+-- | Map the @github@ library's @Error@ type into AppErrorType.
+githubErrToAppErr :: GitHub.Error -> AppErrorType
+githubErrToAppErr err = case err of
+    GitHub.HTTPError httpException -> GitHubError . HTTPError $ show httpException 
+    GitHub.ParseError text         -> GitHubError . ParseError $ text
```
chshersh: I personally don't like `f . g . h $ x` file. I find that it requires to type redundant characters. Something like the following looks simpler to me:

```haskell
GitHubError $ ParseError text
```

Never had problems with this, and I don't know why some people in community force this `. . . $` style :slightly_smiling_face:

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+import qualified Data.Vector as Vector
+
+
+-- | Fetch all repositories built with the Haskell language.
+fetchAllHaskellRepos :: (WithDb env m, WithError m, WithLog env m) => m [GitHub.Repo]
+fetchAllHaskellRepos = liftGitHubSearchToApp searchHaskellRepos 
+
+-- | Fetch all open issues with Haskell language and the labels passed in to the function.
+fetchHaskellIssuesByLabels :: (WithDb env m, WithError m, WithLog env m) => [Label] -> m [GitHub.Issue]
+fetchHaskellIssuesByLabels labels = liftGitHubSearchToApp $ searchHaskellIssuesByLabels labels
```
chshersh:
```suggestion
fetchHaskellIssuesByLabels = liftGitHubSearchToApp . searchHaskellIssuesByLabels
```

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+
+-- | Fetch all open issues with Haskell language.
+fetchAllHaskellIssues :: (WithDb env m, WithError m, WithLog env m) => m [GitHub.Issue]
+fetchAllHaskellIssues = fetchHaskellIssuesByLabels []
+
+-- | Lift a github search action to work within our monad stack.
+liftGitHubSearchToApp :: (WithDb env m, WithError m, WithLog env m) => IO (Either GitHub.Error (SearchResult a)) -> m [a]
```
chshersh: This type signature is too long. I propose to split it into multiple lines according to our style guide.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+liftGitHubSearchToApp githubSearch = do
+    result <- liftIO githubSearch
+    case result of
+        Right (SearchResult count vec) -> do
+            log Info $ "Fetching total of " <> show count <> " results."
```
chshersh: You can add `Typeable`  constraint to `a` type variable and use `typeName @a` here instead of `"results"` to log what data type we are processing now.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+fetchHaskellIssuesByLabels labels = liftGitHubSearchToApp $ searchHaskellIssuesByLabels labels
+
+-- | Fetch all open issues with Haskell language.
+fetchAllHaskellIssues :: (WithDb env m, WithError m, WithLog env m) => m [GitHub.Issue]
+fetchAllHaskellIssues = fetchHaskellIssuesByLabels []
+
+-- | Lift a github search action to work within our monad stack.
+liftGitHubSearchToApp :: (WithDb env m, WithError m, WithLog env m) => IO (Either GitHub.Error (SearchResult a)) -> m [a]
+liftGitHubSearchToApp githubSearch = do
+    result <- liftIO githubSearch
```
chshersh: This is actually a nice use case for `LambdaCase` extension. You can write:

```haskell
liftGitHubSearchToApp searchAction = liftIO searchAction >>= \case
    Left ...
    Right ...
```

This helps to avoid need to come up with the variable names and make nested indentation less nested.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+    case result of
+        Right (SearchResult count vec) -> do
+            log Info $ "Fetching total of " <> show count <> " results." 
+            return $ Vector.toList vec
```
chshersh: We prefer to use `pure` instead of `return` :slightly_smiling_face:

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+        Left err                       -> throwError . githubErrToAppErr $ err
```
chshersh: It's strange that alignment is too big here... According to our style guide, in this case you shouldn't have such big alignment.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+
+-- | Convert a value of the @GitHub.Issue@ type to a value of our own @Issue@ type.
+githubIssueToIssue :: GitHub.Issue -> Issue
+githubIssueToIssue githubIssue = Issue{..}
+  where  
+    issueRepoOwner = undefined
+    issueRepoName  = undefined
+    issueNumber    = GitHub.unIssueNumber . GitHub.issueNumber $ githubIssue
+    issueTitle     = GitHub.issueTitle githubIssue 
+    issueBody      = maybeTextToText . GitHub.issueBody $ githubIssue
```
chshersh: I appreciate good intention to create pure functions and refactor duplication, but `fromMaybe` is okay here, no need to create separate function
```suggestion
    issueBody      = fromMaybe "" $ GitHub.issueBody githubIssue
```

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,89 @@
...
+    issueLabels    = SqlArray . Vector.toList $ (\GitHub.IssueLabel{..} -> Label $ GitHub.untagName labelName) <$> GitHub.issueLabels githubIssue 
+
+-- | Extract the @RepoOwner@ from a value of type @GitHub.Issue@.
+parseIssueUrlRepoOwner :: GitHub.Issue -> Maybe RepoOwner
+parseIssueUrlRepoOwner issue = RepoOwner <$> (listToMaybe . take 1 . drop 4 . split ('/'==) . GitHub.getUrl . GitHub.issueUrl) issue
```
chshersh: I propose to use `stripPrefix` function to get owner and repo. See example from `hit`:

https://github.com/kowainik/hit-on/blob/50c77c051b0e190b1279fac5dcebf34358a09755/src/Hit/Issue.hs#L154-L158

Also, let's have a single function:

```haskell
parseIssueUserData :: Url -> Maybe (RepoOwner, RepoName)
```

src/IW/Sync/Search.hs
rashadg1030: I must’ve read the style guide wrong. Will go back and fix.

src/IW/Sync/Search.hs
rashadg1030: This is a habit. Sorry. I really like the idea of using pure instead of return though. Makes more sense 💡

src/IW/Sync/Search.hs
rashadg1030: Okay good I was thinking the same. Will fix this.

src/IW/App/Error.hs
rashadg1030: Got it👍🏽

src/IW/App/Error.hs
```diff
@@ -79,7 +81,8 @@ data AppError = AppError
     } deriving (Show, Eq)
 
 -- | App errors type.
-newtype AppErrorType = InternalError IError
+data AppErrorType = InternalError IError
+                  | GitHubError GError
```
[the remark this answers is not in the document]
rashadg1030: Right, got it.

issue-wanted.cabal
```diff
@@ -108,8 +112,10 @@ library
                      , servant-server ^>= 0.16
                      , text
                      , time >= 1.8 && < 1.10
+                     , time-compat >=1.9.2.2 && <1.10
```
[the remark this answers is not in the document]
rashadg1030: yeah I think I can remove these I will check.

src/IW/Sync/Search.hs
rashadg1030: Okay cool I was wondering how to do that. Thank you.

rashadg1030:
```
/home/rashad/Desktop/issue-wanted/src/IW/Sync/Search.hs:66:63: error:
    • Cannot apply expression of type ‘t1’
      to a visible type argument ‘a’
    • In the first argument of ‘(<>)’, namely ‘(typeName @a)’
      In the second argument of ‘(<>)’, namely ‘(typeName @a) <> "s..."’
      In the second argument of ‘(<>)’, namely
        ‘show count <> (typeName @a) <> "s..."’
   |        
66 |             log Info $ "Fetching total of " <> show count <> (typeName @a) <> "s..." 
   |                                                               ^^^^^^^^^^^
            
Progress 1/2
```
I'm getting this error when doing `typeName @a`. Can't figure out why. Am I missing a language extension? I tried adding `ExplicitForAll` but nothing happened.

src/IW/Sync/Search.hs
rashadg1030: Nice!

src/IW/Sync/Search.hs
rashadg1030: Awesome :+1:

rashadg1030:
@chshersh Made the requested changes. This PR is pretty much done except for the one question I have above. Also, let me know if you have any name suggestions for the functions and variables.

src/IW/Sync/Search.hs
chshersh: You need to add `forall` explicitly to the function signature. The extension is `ScopedTypeVariables`. It's enabled by default in this project.

chshersh:
@rashadg1030 I've answered your question 🙂 Will review your PR.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,116 @@
...
+    issueNumber = GitHub.unIssueNumber $ GitHub.issueNumber githubIssue
+    issueTitle  = GitHub.issueTitle githubIssue 
+    issueBody   = fromMaybe "" $ GitHub.issueBody githubIssue
+    issueLabels = SqlArray . V.toList $ (\GitHub.IssueLabel{..} -> Label $ GitHub.untagName labelName) <$> GitHub.issueLabels githubIssue 
+
+-- | For parsing the @issueRepoOwner@ and @issueRepoName@ from the issue's URL.
+parseIssueUserData :: GitHub.URL -> Maybe (RepoOwner, RepoName)
+parseIssueUserData (URL url) =
+    ( T.stripPrefix "git@github.com:"     url
+  <|> T.stripPrefix "https://github.com/" url
```
chshersh: I recall you said that the URL for issue (that comes from GitHub) looks like this:

```
https://api.github.com/repos/<repoOwner>/<repoName>/issues/<number>
```

Should be in that case it be `T.stripPrefix "https://api.github.com/repos/"`?

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,116 @@
...
+githubIssueToIssue githubIssue = do
+    issueRepoOwner <- fst <$> parseIssueUserData issueUrl
+    issueRepoName  <- snd <$> parseIssueUserData issueUrl
+    pure Issue{..}
+  where  
+    issueUrl    = GitHub.issueUrl githubIssue
+    issueNumber = GitHub.unIssueNumber $ GitHub.issueNumber githubIssue
+    issueTitle  = GitHub.issueTitle githubIssue 
+    issueBody   = fromMaybe "" $ GitHub.issueBody githubIssue
+    issueLabels = SqlArray . V.toList $ (\GitHub.IssueLabel{..} -> Label $ GitHub.untagName labelName) <$> GitHub.issueLabels githubIssue
```
chshersh: This line is a bit long, probably makes sense to split it into multiple lines

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,116 @@
...
+labelsToSearchQuery :: [Label] -> Text
+labelsToSearchQuery = foldMap (\Label{..} -> "label:\"" <> unLabel <> "\" ")
+
+-- | Search for all open Haskell issues with the corresponding labels.
+searchHaskellIssuesByLabels :: [Label] -> IO (Either GitHub.Error (SearchResult GitHub.Issue))
+searchHaskellIssuesByLabels labels = searchIssues $ "language:haskell is:open " <> labelsToSearchQuery labels
+
+-- | Convert a value of the @GitHub.Repo@ type to a value of our own @Repo@ type.
+githubRepoToRepo :: GitHub.Repo -> Repo
+githubRepoToRepo githubRepo = Repo{..}
```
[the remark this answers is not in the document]
rashadg1030: Got it :+1:

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,116 @@
...
+    repoName       = RepoName  . GitHub.untagName $ GitHub.repoName githubRepo
+    repoDescr      = fromMaybe "" $ GitHub.repoDescription githubRepo
+    repoCategories = SqlArray []
+
+-- | Convert a value of the @GitHub.Issue@ type to a value of our own @Issue@ type.
+githubIssueToIssue :: GitHub.Issue -> Maybe Issue
+githubIssueToIssue githubIssue = do
+    issueRepoOwner <- fst <$> parseIssueUserData issueUrl
+    issueRepoName  <- snd <$> parseIssueUserData issueUrl
+    pure Issue{..}
```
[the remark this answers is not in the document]
rashadg1030: Cool. You can do a lot of things with the record syntax.

src/IW/Sync/Search.hs
rashadg1030: @chshersh Ooops yes that is correct!

rashadg1030: @chshersh And I don't have to worry about a `git@github.com:` prefix right?

chshersh: @rashadg1030 Right, you don't need to worry about alternative prefixes, we have only single prefix.

test/Spec.hs
```diff
@@ -15,13 +15,15 @@ import Test.Common (joinSpecs)
...
 import Test.Db (dbSpecs)
+import Test.Sync (syncSpecs)
 
 import qualified Data.Pool as Pool
 
 
 hspecTests :: AppEnv -> Spec
 hspecTests = sequential . joinSpecs "issue-wanted" 
     [ dbSpecs
+    , syncSpecs
```
chshersh: When you don't need `AppEnv` in some of your tests, it's better to write `const syncSpecs` instead of `_` and specifying argument type.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,118 @@
...
+    >>= stripGitSuffix 
+    >>= splitOwnerAndName
+  where
+    splitOwnerAndName :: Text -> Maybe (RepoOwner, RepoName)
+    splitOwnerAndName strippedUrl =
+        let (owner, name) = fst . T.breakOn "/" . T.drop 1 <$> T.breakOn "/" strippedUrl 
+        in guard (owner /= "" && name /= "") *> Just (RepoOwner owner, RepoName name)
+
+    stripGitSuffix :: Text -> Maybe Text
+    stripGitSuffix x = whenNothing (T.stripSuffix ".git" x) (Just x)
```
chshersh: You don't need to strip `.git` suffix. I guess URL doesn't end with `.git`

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,118 @@
...
+        , issueTitle  = GitHub.issueTitle githubIssue 
+        , issueBody   = fromMaybe "" $ GitHub.issueBody githubIssue
+        , issueLabels = SqlArray 
+              $ V.toList 
+              $ (\GitHub.IssueLabel{..} -> Label $ GitHub.untagName labelName) 
+              <$> GitHub.issueLabels githubIssue
+        , ..
+        }
+
+-- | For parsing the @issueRepoOwner@ and @issueRepoName@ from the issue's URL.
```
chshersh: Could you please add to the documentation the shape or the URL this function parses and the expected output from this URL? This will help to understand the code better

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,118 @@
...
+parseIssueUserData :: GitHub.URL -> Maybe (RepoOwner, RepoName)
+parseIssueUserData (URL url) =
+    T.stripPrefix "https://api.github.com/repos/" url 
+    >>= stripGitSuffix 
+    >>= splitOwnerAndName
+  where
+    splitOwnerAndName :: Text -> Maybe (RepoOwner, RepoName)
+    splitOwnerAndName strippedUrl =
+        let (owner, name) = fst . T.breakOn "/" . T.drop 1 <$> T.breakOn "/" strippedUrl
```
chshersh: I would propose to use `splitOn` function instead of double `breakOn`:

* https://hackage.haskell.org/package/text-1.2.3.1/docs/Data-Text.html#v:splitOn

test/Test/Sync.hs
```diff
@@ -0,0 +1,26 @@
...
+syncSpecs :: AppEnv -> Spec
+syncSpecs = joinSpecs "GitHub sync correctness"
+    [ parseIssueUserDataSpec
+    ]
+
+parseIssueUserDataSpec :: AppEnv -> Spec
+parseIssueUserDataSpec _ = describe "parseIssueUserData" $
+    it "should return Just (RepoOwner owner123, RepoName repo123)" $ 
+        parseIssueUserData testGitHubIssueUrl 
+        `shouldBe` Just (RepoOwner "owner123", RepoName "repo123")
```
chshersh: Very good idea to write tests for this function! Just a minor indentation suggestion for consistency with the style guide
```suggestion
            `shouldBe` Just (RepoOwner "owner123", RepoName "repo123")
```

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,118 @@
...
+fromgitHubIssue :: GitHub.Issue -> Maybe Issue
+fromgitHubIssue githubIssue = do
+    (issueRepoOwner, issueRepoName) <- parseIssueUserData $ GitHub.issueUrl githubIssue
+    pure Issue 
+        { issueNumber = GitHub.unIssueNumber $ GitHub.issueNumber githubIssue
+        , issueTitle  = GitHub.issueTitle githubIssue 
+        , issueBody   = fromMaybe "" $ GitHub.issueBody githubIssue
+        , issueLabels = SqlArray 
+              $ V.toList 
+              $ (\GitHub.IssueLabel{..} -> Label $ GitHub.untagName labelName)
```
chshersh: I think point-free style here looks a bit simpler
```suggestion
              $ (Label . GitHub.untagName . labelName) 
```

rashadg1030:
@chshersh I had something come to mind:
We're using this function, but it is a partial function. If you pass an empty string `""` it throws an error.
```
splitOn :: Text -> Text -> [Text] 
```
This function is safe, but is more code when you use it:
```
split :: (Char -> Bool) -> Text -> [Text]
```
I heard it is bad practice to use partial functions, but in our case:
```
 T.splitOn "/" strippedUrl
```
The value is hardcoded so it is fine. We know for sure it is not an empty string. Thoughts?

chshersh:
@rashadg1030 I think it's okay to use `splitOn` in our case because we know that the argument is not empty. You should use partial functions carefully if you pass runtime values as an argument. For example, somebody gives you a list and you're trying to call `head` function over this list. But since we are in control of what we are doing, it's okay to use a shorter function.

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,117 @@
...
+fromGitHubRepo :: GitHub.Repo -> Repo
+fromGitHubRepo githubRepo = Repo
+    { repoOwner      = RepoOwner $ GitHub.untagName $ GitHub.simpleOwnerLogin $ GitHub.repoOwner githubRepo
+    , repoName       = RepoName  $ GitHub.untagName $ GitHub.repoName githubRepo
+    , repoDescr      = fromMaybe "" $ GitHub.repoDescription githubRepo
+    , repoCategories = SqlArray []
+    }
+
+-- | Convert a value of the @GitHub.Issue@ type to a value of our own @Issue@ type.
+fromgitHubIssue :: GitHub.Issue -> Maybe Issue
```
vrom911: Probably makes sense to name it `fromGitHubIssue` (with capitalized G)?

test/Test/Sync.hs
```diff
@@ -0,0 +1,23 @@
...
+
+syncSpecs :: Spec
+syncSpecs = describe "GitHub sync correctness"
+    parseIssueUserDataSpec
+
+parseIssueUserDataSpec :: Spec
+parseIssueUserDataSpec = describe "parseIssueUserData" $
+    it "should return Just (RepoOwner owner123, RepoName repo123)" $ 
+        parseIssueUserData testGitHubIssueUrl 
+            `shouldBe` Just (RepoOwner "owner123", RepoName "repo123")
```
vrom911: Don't you mind to add a few more tests on wrong URLs, just to check that it doesn't parse wrong input?

src/IW/Sync/Search.hs
```diff
@@ -0,0 +1,117 @@
+module IW.Sync.Search
```
vrom911: Could we add documentation to this whole module as well so that it would be even clearer what are these endpoints and when they are going to be used? This could be useful when you looking quickly through the modules of the project.

```
{- | This module provides functions...
-}
```
What do you think?

src/IW/Sync/Search.hs
rashadg1030: @vrom911 Definitely should be capital G.

src/IW/Sync/Search.hs
rashadg1030: @vrom911 I agree! I need to add some documentation to some other modules as well.

test/Test/Sync.hs
rashadg1030: @vrom911 Good idea💡

rashadg1030:
@vrom911 @chshersh Made suggested changes. It doesn't pass on a test that I feel like it should pass, but check it out and let me know what you think.

chshersh:
@rashadg1030 I think current tests are okay. URL for GitHub won't change that often, so we don't need to handle all various cases that won't happen in practice.

test/Test/Sync.hs
```diff
@@ -0,0 +1,41 @@
...
+syncSpecs :: Spec
+syncSpecs = describe "GitHub sync correctness"
+    parseIssueUserDataSpec
+
+parseIssueUserDataSpec :: Spec
+parseIssueUserDataSpec = describe "parseIssueUserData" $ do
+    it "should return Just (RepoOwner owner123, RepoName repo123)" $ 
+        parseIssueUserData testGitHubIssueUrl 
+            `shouldBe` Just (RepoOwner "owner123", RepoName "repo123")
+    it "should return Nothing" $
```
chshersh: Let's also add the name of the parsed URL to the test, so we know which exact tests is failing. Having test cases under the same name under `it` is a bad practice, because tests can perform concurrently or some output might be interleaved. So all test cases should have unique names.

test/Test/Sync.hs
```diff
@@ -0,0 +1,41 @@
...
+testGitHubIssueUrl = GitHub.URL "https://api.github.com/repos/owner123/repo123/issues/1"
+
+testBadGitHubIssueUrl1 :: GitHub.URL
+testBadGitHubIssueUrl1 = GitHub.URL "https://api.github.com/repos/owner123/issues/1"
+
+testBadGitHubIssueUrl2 :: GitHub.URL
+testBadGitHubIssueUrl2 = GitHub.URL "https://api.github.com/randomWord/owner123/repo123/issues/1"
+
+testBadGitHubIssueUrl3 :: GitHub.URL
+testBadGitHubIssueUrl3 = GitHub.URL "api.github.com/repos/owner123/repo123/issues/1"
```
chshersh: Usually we put test data at the end of the file, so that exported specs are shown immediately. Structure of the files should follow `where`-style of definitions.

test/Test/Sync.hs
```diff
@@ -0,0 +1,41 @@
...
+import Test.Hspec (Spec, describe, it, shouldBe)
+
+import qualified GitHub  
+
+
+testGitHubIssueUrl :: GitHub.URL
+testGitHubIssueUrl = GitHub.URL "https://api.github.com/repos/owner123/repo123/issues/1"
+
+testBadGitHubIssueUrl1 :: GitHub.URL
+testBadGitHubIssueUrl1 = GitHub.URL "https://api.github.com/repos/owner123/issues/1"
```
chshersh: This URL looks exactly like `testGitHubIssueUrl` that's why tests are failing currently

rashadg1030: Well the `testBadGitHubIssueUrl1` is missing the `repoName` `repo123` in the URL.

rashadg1030: I''ll remove it for now.

chshersh: > I feel that Hakyll is more of a monolithic choice when it comes to static site generators

Ah, I see. Well, in this case there's nothing we can do, be need extremely strict URL parser to handle such cases. But, again, we expect GitHub to be fair.

PR #108, opened 2019-07-10
[#107] Add Url newtype for wrapping URL text

### LINKED ISSUE #107: Add `Url` newtype for wrapping URL text
`newtype Url = Url { unUrl :: Text }`

rashadg1030:
Resolves #107

### DIFF
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -8,12 +8,13 @@ module IW.Core.Issue
   
 import IW.Core.SqlArray (SqlArray (..))
 import IW.Core.Repo (RepoName (..), RepoOwner (..))
+import IW.Core.Url (Url (..))
 
 
 -- | Wrapper for issue label names.
 newtype Label = Label { unLabel :: Text }
-    deriving stock   (Generic, Show)
-    deriving newtype (Eq, Ord, FromField, ToField, FromJSON, ToJSON, FromHttpApiData)
+    deriving stock   (Generic)
+    deriving newtype (Eq, Ord, Show, FromField, ToField, FromJSON, ToJSON, FromHttpApiData)
 
 -- | Data type representing a GitHub issue.
 data Issue = Issue
@@ -26,12 +27,12 @@ data Issue = Issue
     } deriving stock (Eq, Generic, Show)
       deriving anyclass (ToJSON, FromRow, ToRow)
 
-issueUrl :: Issue -> Text
-issueUrl Issue{..} = 
-    "https://github.com/" 
+issueUrl :: Issue -> Url
+issueUrl Issue{..} = Url  
+    $ "https://github.com/" 
     <> unRepoOwner issueRepoOwner 
     <> "/" 
     <> unRepoName issueRepoName 
     <> "/issues/" 
     <> show issueNumber
-    
+ 
--- a/src/IW/Core/Issue.hs
+++ b/src/IW/Core/Issue.hs
@@ -13,8 +13,8 @@ import IW.Core.Url (Url (..))
 
 -- | Wrapper for issue label names.
 newtype Label = Label { unLabel :: Text }
-    deriving stock   (Generic)
-    deriving newtype (Eq, Ord, Show, FromField, ToField, FromJSON, ToJSON, FromHttpApiData)
+    deriving stock   (Generic, Show)
+    deriving newtype (Eq, Ord, FromField, ToField, FromJSON, ToJSON, FromHttpApiData)
 
 -- | Data type representing a GitHub issue.
 data Issue = Issue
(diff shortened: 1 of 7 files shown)


src/IW/Core/Url.hs
```diff
@@ -0,0 +1,8 @@
+module IW.Core.Url
+       ( Url (..)
+       ) where
+
+
+newtype Url = Url { unUrl :: Text }
```
vrom911: By style guide we do:

```haskell
newtype Url = Url 
    { unUrl :: Text 
    } deriving ...
```

src/IW/Core/Url.hs
```diff
@@ -0,0 +1,8 @@
...
+    deriving stock   (Generic, Show)
```
vrom911: Oh, also, I've just thought, maybe it's better to derive `Show` with `newtype` strategy? It would look more straightforward :thinking:

rashadg1030: @vrom911 I'm not really familiar with deriving strategies, but that sounds good to me. `Generic` as well?

src/IW/Core/Url.hs
rashadg1030: @vrom911 Yep, sorry completely overlooked that for some reason.

src/IW/Core/Url.hs
vrom911: Deriving in `Show` case is quite visible and it can be used as a good example to display the difference between strategies. When you derive with the `stock` strategy (this is the default one and equvivalent to just `deriving Show`) then the standard instance will be generated for the data type. In case of newtypes it would look like this:

```haskell
ghci> show $ Url "abc"
Url { unUrl = "abc" }
```

`newtype` deriving strategy would produce the same instances like the ones on the underlying type (in our case as `Text` type instances). And in this case it would be:

```haskell
ghci> show $ Url "abc"
"abc"
```

Unfortunately, I don't know any good links to provide to read about this better :( Probably @chshersh could help me here.

`Generic` should be derived with `stock` strategy.

rashadg1030: @vrom911 You explained this nicely. Thank you.

chshersh:
Okay, now this PR can be updated :slightly_smiling_face:

rashadg1030:
@chshersh Got it!

rashadg1030:
@chshersh Rebased. Should be good to go.

src/IW/Core/Issue.hs
```diff
@@ -8,12 +8,13 @@ module IW.Core.Issue
...
 import IW.Core.Repo (RepoName (..), RepoOwner (..))
+import IW.Core.Url (Url (..))
 
 
 -- | Wrapper for issue label names.
 newtype Label = Label { unLabel :: Text }
-    deriving stock   (Generic, Show)
-    deriving newtype (Eq, Ord, FromField, ToField, FromJSON, ToJSON, FromHttpApiData)
+    deriving stock   (Generic)
+    deriving newtype (Eq, Ord, Show, FromField, ToField, FromJSON, ToJSON, FromHttpApiData)
```
chshersh: I'm not sure whether @vrom911 meant to change every `Show` deriving or just for `Url`... :thinking:

rashadg1030: @chshersh probably just URL

rashadg1030:
> Also, probably we can use `Url` in `IW.Sync.Search` module. What do you think?
`import GitHub (SearchResult (..), URL (..))`
We use the `GitHub.URL` so I don't see a need for our `URL` type.
```
parseIssueUserData :: GitHub.URL -> Maybe (RepoOwner, RepoName)
```
The function above needs to use the `GitHub.URL`.

rashadg1030:
There is a warning here I don't think CI catches it:
```
let owner:name:_ = T.splitOn "/" strippedUrl
```
```
warning: [-Wincomplete-uni-patterns]
    Pattern match(es) are non-exhaustive
```

chshersh:
> We use the GitHub.URL so I don't see a need for our URL type.

Fair point. We can localize GitHub data types in a single module, makes sense.

> There is a warning here I don't think CI catches it:

We don't build with `-Werror` on CI and rely on people fixing warnings. Regarding this one: it's a bit annoying. I propose to fix it like this:

```haskell
splitOwnerAndName :: Text -> Maybe (RepoOwner, RepoName)
splitOwnerAndName strippedUrl = do  -- do-notation for Maybe monad
    owner:name:_ <- Just $ T.splitOn "/" strippedUrl  -- this will result to Nothing in failed case due to MonadFail instance
    guard $ owner /= "" && name /= ""
    pure (RepoOwner owner, RepoName name)
```

rashadg1030:
@chshersh @vrom911 Made suggested changes.

PR #109, opened 2019-07-10
[#105] Implement `MonadDownload` effect

### LINKED ISSUE #105: Implement `MonadDownload` effect
This effect should take `RepoOwner` and `RepoName`, construct a URL for corresponding `.cabal` file, try to download the file and throw the error in case of unsuccess.

This involves multiple steps:

1. Create `IW.Effects.Download` module
2. Introduce the following typeclass there:
```haskell
class Monad m => MonadDownload m where
    downloadFile :: Text -> m ByteString  -- probably better type for URL can be used instead of Text
```
3. Add `Manager` field to `Env` data type
4. Implement `Has` instance for `Env` and `Manager`
5. Implement `MonadDownload` for `App` in terms of `Has Manager` and `WithError m`

See implementation details here:

* https://github.com/kowainik/issue-wanted/issues/43#issuecomment-509884577

rashadg1030:
Resolves #105

### DIFF
--- a/src/IW.hs
+++ b/src/IW.hs
@@ -5,6 +5,7 @@ module IW
        , main
        ) where
 
+import Network.HTTP.Client.TLS (newTlsManager)
 import Network.Wai.Handler.Warp (run)
 import Servant.Server (serve)
 
@@ -19,6 +20,9 @@ mkAppEnv :: Config -> IO AppEnv
 mkAppEnv Config{..} = do
     -- IO configuration
     envDbPool <- initialisePool cDbCredentials
+
+    -- Http manager configuration
+    envManager <- newTlsManager
 
     -- pure configuration
     let envLogAction = mainLogAction cLogSeverity
--- a/src/IW/Effects/Download.hs
+++ b/src/IW/Effects/Download.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE AllowAmbiguousTypes  #-}
-
 {- | This module contains the class definitiion of @MonadDownload@ and 
 an instance of @MonadDownload@ for the @App@ monad. Instances of 
 @MonadDownload@ have a @downloadFile@ action for downloading files from a given URL.
--- a/src/IW/Effects/Download.hs
+++ b/src/IW/Effects/Download.hs
@@ -1,0 +1,38 @@
+{-# LANGUAGE AllowAmbiguousTypes  #-}
+
+{- | This module contains the class definitiion of @MonadDownload@ and 
+an instance of @MonadDownload@ for the @App@ monad. Instances of 
+@MonadDownload@ have a @downloadFile@ action for downloading files from a given URL.
+-}
+
+module IW.Effects.Download
+       ( MonadDownload (..)
+        
+       -- * Internals
+       , downloadFileImpl
+       ) where
+
+import Network.HTTP.Client (Manager, Response (..), httpLbs)
+import Network.HTTP.Types (Status (..))
+
+import IW.App (App, Has, WithError, grab, throwError, notFound)   
+import IW.Core.Url (Url (..)) 
+
+
+-- | Describes a monad that can download files from a given URL.
+class Monad m => MonadDownload m where
+    downloadFile :: Url -> m ByteString
+
+instance MonadDownload App where 
+    downloadFile = downloadFileImpl
+
+type WithDownload env m = (MonadIO m, MonadReader env m, WithError m, Has Manager env)
+
+downloadFileImpl :: WithDownload env m => Url -> m ByteString
+downloadFileImpl Url{..} = do
+    man <- grab @Manager
+    let req = fromString $ toString unUrl
+    response <- liftIO $ httpLbs req man
+    case statusCode $ responseStatus response of
+        200 -> pure $ toStrict $ responseBody response
+        _   -> throwError notFound
--- a/test/Test/Sync.hs
+++ b/test/Test/Sync.hs
@@ -9,7 +9,7 @@ import IW.Core.Repo (RepoOwner (..), RepoName (..))
 import IW.Core.Url (Url (..))
 import IW.Effects.Download (downloadFileImpl)
 import IW.Sync.Search (parseIssueUserData)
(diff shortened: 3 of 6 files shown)


src/IW/Effects/Download.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE AllowAmbiguousTypes  #-}
+
+module IW.Effects.Download where
```
vrom911: It's better to write the list of exports in each module. Here the list is not going to be huge anyway :slightly_smiling_face:

src/IW/Effects/Download.hs
```diff
@@ -0,0 +1,27 @@
...
+type WithDownload env m = (MonadIO m, MonadReader env m, WithError m, Has Manager env)
+
+downloadFileImpl :: WithDownload env m => Url -> m ByteString
+downloadFileImpl Url{..} = do
+    man <- grab @Manager
+    let req = fromString $ toString unUrl
+    response <- liftIO $ httpLbs req man
+    pure $ case statusCode $ responseStatus response of
+        200       -> toStrict $ responseBody response
+        otherCode -> undefined
```
vrom911: We should use `throwError` function in here instead.

src/IW/Effects/Download.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE AllowAmbiguousTypes  #-}
+
+module IW.Effects.Download where
+
+import Network.HTTP.Client (Manager, Request (..), Response (..), httpLbs)
+import Network.HTTP.Types (Status (..))
+
+import IW.App (App (..), grab, Has, WithError)
```
vrom911: Is `App (..)` necessary here? Wouldn't just `App` do?

src/IW/Effects/Download.hs
```diff
@@ -0,0 +1,27 @@
...
+module IW.Effects.Download where
+
+import Network.HTTP.Client (Manager, Request (..), Response (..), httpLbs)
+import Network.HTTP.Types (Status (..))
+
+import IW.App (App (..), grab, Has, WithError)   
+import IW.Core.Url (Url (..)) 
+
+
+class Monad m => MonadDownload m where
```
vrom911: Let's add some documentation here or at the top of the module :slightly_smiling_face:

src/IW/Effects/Download.hs
rashadg1030: @vrom911 Yep, App is just fine.

src/IW/Effects/Download.hs
rashadg1030: @vrom911 For sure :+1:

src/IW/Effects/Download.hs
rashadg1030: @vrom911 Yep will add that. I guess for now we can just have a simple text error, but in a separate issue I propose to make types for download errors. Here #110

vrom911: Oh, I see! Thanks for pointing that out :slightly_smiling_face:

src/IW.hs
```diff
@@ -20,6 +21,9 @@ mkAppEnv Config{..} = do
     -- IO configuration
     envDbPool <- initialisePool cDbCredentials
 
+    -- Http manager configuration
+    envManager <- newManager defaultManagerSettings
```
chshersh: You should use `newTlsManager` function from `http-client-tls` package here to be able to download from HTTPS urls

src/IW/Effects/Download.hs
```diff
@@ -0,0 +1,37 @@
...
+
+import IW.App (App, Has, WithError, grab, throwError, notFound)   
+import IW.Core.Url (Url (..)) 
+
+-- | Describes a monad that can download files from a given URL.
+class Monad m => MonadDownload m where
+    downloadFile :: Url -> m ByteString
+
+instance MonadDownload App where 
+    downloadFile = downloadFileImpl
```
chshersh: I think it should be possible to write couple tests for this function. Our tests should be able to download `.cabal` from `issue-wanted` itself, I guess

src/IW.hs
rashadg1030: @chshersh Got it!

src/IW/Effects/Download.hs
rashadg1030: @chshersh Sounds good :+1:

rashadg1030:
@chshersh @vrom911 Made suggested changes.

rashadg1030:
@chshersh Rebased on master.

src/IW/Effects/Download.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE AllowAmbiguousTypes  #-}
```
chshersh: I wonder, does this extension is needed here?

test/Test/Sync.hs
```diff
@@ -33,3 +39,16 @@ testBadGitHubIssueUrl1 = GitHub.URL "https://api.github.com/randomWord/owner123/
 
 testBadGitHubIssueUrl2 :: GitHub.URL
 testBadGitHubIssueUrl2 = GitHub.URL "api.github.com/repos/owner123/repo123/issues/1"
+
+downloadFileSpec :: AppEnv -> Spec
+downloadFileSpec env = describe "downloadFile" $ do
+    it "should succeed with 200 status code when passed a valid Url" $
+       env & succeeds (downloadFileImpl issueWantedUrl)
+    it "should fail with notFound error when passed a non-existent Url" $
```
chshersh: Let's add one more test using `equals` to compare the content with some file. Just in case, to make sure, that we actually get the proper content 🙂 The following file is the smallest in our repo that I've found:

* https://github.com/kowainik/issue-wanted/blob/master/app/Main.hs

src/IW/Effects/Download.hs
rashadg1030: @chshersh I was testing something out and forgot to remove it. It's unnecessary now

rashadg1030:
@chshersh Made suggested changes

Issue #110, opened 2019-07-10
Remove Internal/External error wrappers used in `AppErrorType`


rashadg1030:
I think we will need a new constructor for `AppErrorType` for errors that might be thrown by the `MonadDownload` effect.
```
newtype AppErrorType = DownloadError DError
    deriving (Show, Eq)
```
And also helper function for converting the `Status` type in https://hackage.haskell.org/package/http-types-0.12.3/docs/Network-HTTP-Types-Status.html#t:Status to `AppErrorType`. I think we will need this, but `200` status is not an error so...We also need logging but this is a separate issue.

chshersh:
@rashadg1030 I'm starting to worry about inconsistency of our error data types. I would like to have some rules whether we should add constructors to top-level `AppErrorType` or to one of the constructors of this data type. For example, documentation to `InternalError` says the following:

https://github.com/kowainik/issue-wanted/blob/d614fdfeb11c99f33a054c6854c4886ca4180d43/src/IW/App/Error.hs#L89-L92

I think it makes sense instead of:

```haskell
data AppErrorType 
    = InternalError IError
    | GitHubError GError
```

Have something like that:

```haskell
data AppErrorType 
    = InternalError IError
    | ExternalError EError
```

But now, what is the distinction between `InternalError` and `ExternalError`? And if the distinction is subtle, do we really need it to overheat our brains each time when deciding to which data type we need to add constructors?

In `three-layer` this was done this way because there is a mobile application and we need to tell different errors to the mobile application, it handles it independently. But here it looks like all errors will be handled by us, frontend will just display errors. So why not remove `InternalError` and leave only `AppErrorType` with many constructors?

Any thoughts? @vrom911 @rashadg1030

rashadg1030:
@chshersh I think this is a good idea 💡 It is annoying to have that extra wrapper and I don’t think it’s necessary. In my opinion it’s not really helpful to distinguish between Internal and External errors. All the constructors should probably fall under just AppErrorType.

chshersh:
I think we reached consensus over this issue and this refactoring can be performed 🙂

Issue #102, opened 2019-07-10
Add handling of cabal file parser errors


rashadg1030:
Currently using: 
```
parseGenericPackageDescriptionMaybe :: ByteString -> Maybe GenericPackageDescription 
```

Eventually we should use:
```
parseGenericPackageDescription :: ByteString -> ParseResult GenericPackageDescription 
```
And take advantadge of errors.

Related to #110

rashadg1030:
@chshersh @vrom911 Actually maybe this is better left as `Maybe` to keep things simpler, we do have logging in case it is `Nothing`. Should we throw an error as well? If so, this will stop the program each time it is `Nothing` which happens a lot. I guess we can continue the program using `catchError` when an error is thrown.

chshersh:
@rashadg1030 Logging is fine. But I guess that if you use `Result` instead of `Maybe` you should be able to log a more meaningful message.

PR #119, opened 2019-07-11
[#106] Implement MonadCabal effect

### LINKED ISSUE #43: Retrieve category from repository Cabal file
We can use the GitHub API to retrieve the contents of a repo's Cabal file. For example pandoc: https://api.github.com/repos/jgm/pandoc/contents/pandoc.cabal

We can then use 
https://www.haskell.org/cabal/release/cabal-latest/doc/API/Cabal/Distribution-PackageDescription-Parsec.html
to parse the file once we retrieve it.

Anything I'm missing anything or suggestions??

### LINKED ISSUE #103: Write test for function that generates cabal file URL from Repo
Write test for this function once #95 is merged:
`repoCabalFileUrl :: Repo -> Url`

### LINKED ISSUE #106: Implement `MonadCabal` effect for extracting relevant fields of the `.cabal` file
Blocked by #105 

Once `MonadDownload` is implemented, it should be possible to implement:

```haskell
class Monad m => MonadCabal m
    getCabalCategories :: RepoOwner -> RepoName -> m [Category]
```

An instance of this effect for `App` should do the following:

1. Construct URL to `.cabal` file by `RepoOwner` and `RepoName`
2. Download the file.
3. Try to parse it and return a list of categories from this file.

rashadg1030:
Resolves #106 
Resolves #103
Resolves #43


src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,54 @@
...
+repoCabalUrl (RepoOwner rO) (RepoName rN) = Url $
+    "https://raw.githubusercontent.com/" 
+    <> rO
+    <> "/" 
+    <> rN
+    <> "/master/"
+    <> rN
+    <> ".cabal" 
+
+categoryNames :: GenericPackageDescription -> [Category]
```
chshersh: Heh, I didn't expect that `category` is stored as just `String` in Cabal package description. I was expecting that this is already a list...

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,54 @@
...
+categoryNames genPkgDescr = Category . strip <$> splitCategories genPkgDescr
+  where
+    splitCategories :: GenericPackageDescription -> [Text]
+    splitCategories = splitOn' . toText . category . packageDescription
+
+    splitOn' :: Text -> [Text]
```
chshersh: `splitOn'` is not the best name and it's not clear from the name what this function does... Probably it's better to rename and write some comment explaining behaviour. I find it not straightforward to understand.

issue-wanted.cabal
```diff
@@ -95,9 +96,10 @@ library
                          
                          -- Sync Worker
                          IW.Sync.Search  
-
+
```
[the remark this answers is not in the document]
rashadg1030: I'm not sure, I will look into this! I'm sure there is some extension I can use.

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,54 @@
+module IW.Effects.Cabal
```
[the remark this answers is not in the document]
rashadg1030: @chshersh On it. I've been thinking about documentation and I was wondering if it is better to write the docs for the function you're gonna write first and then implement the function itself?

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,54 @@
...
+
+getCabalCategoriesImpl :: WithCabal env m => RepoOwner -> RepoName -> m [Category]
+getCabalCategoriesImpl repoOwner repoName = do
+    cabalFile <- downloadFileImpl $ repoCabalUrl repoOwner repoName 
+    pure $ case parseGenericPackageDescriptionMaybe cabalFile of
+        Nothing -> []
+        Just genPkgDescr -> categoryNames genPkgDescr
+
+repoCabalUrl :: RepoOwner -> RepoName -> Url
+repoCabalUrl (RepoOwner rO) (RepoName rN) = Url $
```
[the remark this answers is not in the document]
rashadg1030: @chshersh Yep, got it :+1:

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,54 @@
...
+
+class Monad m => MonadCabal m where
+    getCabalCategories :: RepoOwner -> RepoName -> m [Category]
+
+instance MonadCabal App where
+    getCabalCategories = getCabalCategoriesImpl
+
+type WithCabal env m = (MonadIO m, MonadReader env m, WithError m, Has Manager env)
+
+getCabalCategoriesImpl :: WithCabal env m => RepoOwner -> RepoName -> m [Category]
```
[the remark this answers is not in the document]
rashadg1030: @chshersh Got it.

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,54 @@
...
+        Nothing -> []
+        Just genPkgDescr -> categoryNames genPkgDescr
+
+repoCabalUrl :: RepoOwner -> RepoName -> Url
+repoCabalUrl (RepoOwner rO) (RepoName rN) = Url $
+    "https://raw.githubusercontent.com/" 
+    <> rO
+    <> "/" 
+    <> rN
+    <> "/master/"
```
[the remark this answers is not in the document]
rashadg1030: @chshersh That’s true. I will see if there’s a way to get the default branch.

src/IW/Effects/Cabal.hs
rashadg1030: @chshersh haha yeah I though this kind of odd as well

src/IW/Effects/Cabal.hs
rashadg1030: @chshersh got it. I will the right name.

src/IW/Effects/Cabal.hs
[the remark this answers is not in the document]
chshersh: @rashadg1030 Hard to tell :thinking: I think this is individual. Usually, I write documentation only after I've implemented a function. Because during implementation I might realize that I don't even need this function at all! So why spend time on writing docs? :slightly_smiling_face: I've never heard of DDD -- documentation-driven development. But I can imagine that for some people it might be easier to describe the intention in natural language and then implement the thing.

test/Test/Sync.hs
```diff
@@ -17,7 +18,8 @@ import qualified GitHub
 syncSpecs :: AppEnv -> Spec
 syncSpecs env = describe "GitHub sync correctness" $ do
     parseIssueUserDataSpec
-    downloadFileSpec env 
+    downloadFileSpec env
```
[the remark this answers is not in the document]
rashadg1030: @chshersh Sounds good. Actually, in another branch, I wrote a test for `repoCabalUrl`. Refer to issue #103. I think I'm just gonna add it to this branch.

src/IW/Db/Repo.hs
```diff
@@ -33,12 +33,12 @@ getReposByCategory = query [sql|
 -- | Insert a list of repos into the database, but update on conflict
 upsertRepos :: (WithDb env m) => [Repo] -> m ()
 upsertRepos = executeMany [sql|
-    INSERT INTO repos 
+    INSERT INTO repos
         (owner, name, descr, categories)
-    VALUES
```
chshersh: Interesting fact, but having trailing whitespaces in SQL queries can actually decrease performance :slightly_smiling_face: Because these queries are parsed at runtime and sent via socket to DB. So, the more characters they have the slower they will be. Well, even if we have 10 trailing whitespaces on each line, we won't even notice the slowdown. It's just an interesting observation of how a style guide can affect performance

rashadg1030: Wow, that is an interesting fact. I had no idea whitespace could make your program slower.

chshersh: And you haven't even heard about stories where having comments in your code can make your code slower...

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,61 @@
+{- | This module contains the class definitiion of @MonadCabal@ and
+an instance of @MonadCabald@ for the @App@ monad. Instances of
+@MonadCabal@ have a @getCabalCategories@ action that returns @[Category]@
+given a @RepoOwner@ and @RepoName@.
```
vrom911: Probably it worth adding that it does so by downloading the `.cabal` file and parsing the Categories field of the file. So it's clearer what to expect from this function.

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,61 @@
...
+
+import IW.App (App)
+import IW.Core.Repo (RepoOwner (..), RepoName (..), Category (..))
+import IW.Core.Url (Url (..))
+import IW.Effects.Download (MonadDownload (..))
+
+
+-- | Describes a monad that returns @[Category]@ given a @RepoOwner@ and @RepoName@.
+class Monad m => MonadCabal m where
+    getCabalCategories :: RepoOwner -> RepoName -> m [Category]
```
vrom911: Maybe we can somehow mention which errors we can expect from this function? (cc @chshersh )
Because it's using `MonadDownload` we can say this explicitly. What do you think?

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,61 @@
...
+
+instance MonadCabal App where
+    getCabalCategories = getCabalCategoriesImpl
+
+getCabalCategoriesImpl :: MonadDownload m => RepoOwner -> RepoName -> m [Category]
+getCabalCategoriesImpl repoOwner repoName = do
+    cabalFile <- downloadFile $ repoCabalUrl repoOwner repoName
+    pure $ case parseGenericPackageDescriptionMaybe cabalFile of
```
vrom911: Let's add debugging log message that we are trying to parse *owner/repo/name.cabal* file now.

Also I assume that there is some logging about download, so we don't need to add logs here as well :slightly_smiling_face:

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,61 @@
...
+        Nothing -> []
```
vrom911: Could we add log that parsing failed? This is an interesting case for us to look into in the future.

src/IW/Effects/Download.hs
```diff
@@ -1,27 +1,27 @@
...
 
 
--- | Describes a monad that can download files from a given URL.
+-- | Describes a monad that can download files from a given @Url@.
 class Monad m => MonadDownload m where
     downloadFile :: Url -> m ByteString
 
-instance MonadDownload App where 
+instance MonadDownload App where
     downloadFile = downloadFileImpl
```
vrom911: Oh, I see now that the logging is not added here yet.

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 Sounds good. Should I just add logging and errors to this PR? I could do it right now

src/IW/Effects/Download.hs
rashadg1030: @vrom911 I could add logging to this PR. But I think it is kinda a lot because I will have to write functions to help convert other library errors to our own, and I have to change our `AppErrorType`:
refer to #110. We're still undecided there I think. Perhaps you could provide your thoughts?

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 Okay good I was wondering if we needed logging for both, but logging for `MonadDownload` should be enough, right?

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 I think for now I'll just add comments specifying the types of errors that may be thrown.

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 Good idea.

src/IW/Effects/Cabal.hs
chshersh: It would be nice to specify errors on type-level. Though, I'm not aware of a convenient way to do so... I guess having just a comment should work for now :+1:

rashadg1030: @chshersh yes for sure. I will add types for errors as soon as we resolve this discussion: 
#110

rashadg1030: @vrom911 @chshersh The errors for parsing the cabal package description are defined here:
https://hackage.haskell.org/package/Cabal-2.4.1.0/docs/Distribution-PackageDescription-Parsec.html#t:ParseResult
How do we integrate this into our code? For now, we are using the @Maybe@ version of the parse function which returns @Nothing@ on unsuccessful parse, but we want helpful error right?

src/IW/Effects/Download.hs
rashadg1030: @vrom911 I added logging to `MonadDownload1 as well.

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 Added.

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 @chshersh Also refer to #102 for more discussion.

chshersh: @rashadg1030 As I understand, #102 is about using something similar to `Either` instead of `Maybe` to tell specific parsing error. For now we can just tell that file failed to parse and improve the text of the message later.

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,75 @@
+{- | This module contains the class definitiion of @MonadCabal@ and
+an instance of @MonadCabal@ for the @App@ monad. Instances of
+@MonadCabal@ have a @getCabalCategories@ action that returns @[Category]@
+given a @RepoOwner@ and @RepoName@. It does so by downloading a @.cabal@ file
+and parsing the @category@ field of the file.
+-}
+
+module IW.Effects.Cabal
+       ( getCabalCategories
```
chshersh: Usually, we export the whole effect with all its methods
```suggestion
       ( MonadCabal (..)
```

src/IW/Effects/Cabal.hs
```diff
@@ -0,0 +1,75 @@
...
+    case parseGenericPackageDescriptionMaybe cabalFile of
+        Nothing -> do
+            log E $ "Couldn't parse file downloaded from " <> textUrl
+            pure []
+        Just genPkgDescr -> do
+            log I $ "Successfully parsed file downloaded from " <> textUrl
+            pure $ categoryNames genPkgDescr
+  where
+    textUrl :: Text
+    textUrl = unUrl $ repoCabalUrl repoOwner repoName
```
chshersh: `repoCabalUrl repoOwner repoName` duplicates twice: in `where` and inside the function.

src/IW/Effects/Download.hs
```diff
@@ -1,36 +1,41 @@
...
+    log I $ "Attempting to download file from " <> unUrl <> "..."
     response <- liftIO $ httpLbs req man
     case statusCode $ responseStatus response of
-        200 -> pure $ toStrict $ responseBody response
-        _   -> throwError notFound
+        200 -> do
+            log I $ "Successfully downloaded file from " <> unUrl
+            pure $ toStrict $ responseBody response
+        _   -> do
+            log E $ "Couldn't download file from " <> unUrl
```
chshersh: You can also add to logging message the resulting status code that you received. So if we know, whether it's 404 and there's no such file or 500 (some problems with GitHub)

rashadg1030:
@chshersh Made suggested changes.

PR #121, opened 2019-07-12
[#92] Implement better DB filter functions along with their tests

### LINKED ISSUE #92: Improve functionality of filtering DB functions
On the frontend, it should be possible to specify multiple categories and query all repos that have one of the specified categories. Same for labels. This will become relevant once we implement API endpoints and integration with the frontend.

rashadg1030:
Resolves #92

### DIFF
--- a/src/IW/Db/Issue.hs
+++ b/src/IW/Db/Issue.hs
@@ -4,16 +4,17 @@
 
 module IW.Db.Issue
        ( getIssues
-       , getIssuesByLabel
+       , getIssuesByLabels
        , upsertIssues
        ) where
 
-import IW.Core.Issue (Issue (..))
+import IW.Core.Issue (Issue (..), Label (..))
+import IW.Core.SqlArray (SqlArray (..))
 import IW.Core.WithId (WithId)
 import IW.Db.Functions (WithDb, executeMany, query, queryRaw)
 
 
--- | Returns all issues in the database
+-- | Returns all issues in the database.
 getIssues :: (WithDb env m) => m [WithId Issue]
 getIssues = queryRaw [sql|
     SELECT id, repo_owner, repo_name, number, title, body, labels
@@ -21,16 +22,16 @@ getIssues = queryRaw [sql|
     LIMIT 100
 |]
 
--- | Returns a list of issues filtered by label name
-getIssuesByLabel :: (WithDb env m) => Text -> m [WithId Issue]
-getIssuesByLabel = query [sql|
+-- | Returns all issues with at least one label in the given list.
+getIssuesByLabels :: (WithDb env m) => [Label] -> m [WithId Issue]
+getIssuesByLabels = query [sql|
     SELECT id, repo_owner, repo_name, number, title, body, labels
     FROM issues
-    WHERE ? = ANY (labels)
+    WHERE ? && labels
     LIMIT 100
-|] . Only
+|] . Only . SqlArray
 
--- | Insert a list of issues into the database, but update on conflict
+-- | Insert a list of issues into the database, but update on conflict.
 upsertIssues :: (WithDb env m) => [Issue] -> m ()
 upsertIssues = executeMany [sql|
     INSERT INTO issues
--- a/src/IW/Server/Issue.hs
+++ b/src/IW/Server/Issue.hs
@@ -7,16 +7,16 @@ module IW.Server.Issue
        , issuesHandler
        ) where
 
-import IW.Core.Issue (Issue (..))
+import IW.Core.Issue (Issue (..), Label (..))
 import IW.Core.WithId (WithId (..))
-import IW.Db (WithDb, getIssues, getIssuesByLabel)
+import IW.Db (WithDb, getIssuesByLabels)
 import IW.Server.Types (AppServer, ToApi)
 
 
 newtype IssueSite route = IssueSite
     { issuesRoute :: route
         :- "issues"
-        :> QueryParam "label" Text
+        :> ReqBody '[JSON] [Label]
         :> Get '[JSON] [WithId Issue]
     } deriving (Generic)
 
@@ -30,8 +30,6 @@ issueServer = IssueSite
 issuesHandler
     :: ( WithDb env m
        )
-    => Maybe Text
+    => [Label]
     -> m [WithId Issue]
-issuesHandler maybeLabel = case maybeLabel of
-    Nothing    -> getIssues
-    Just label -> getIssuesByLabel label
+issuesHandler = getIssuesByLabels
--- a/src/Prelude.hs
+++ b/src/Prelude.hs
@@ -12,6 +12,7 @@ module Prelude
 
        , WithLog
        , typeName
+       , splitAndStrip
        ) where
 
 -- Reexport
@@ -23,6 +24,8 @@ import Control.Lens ((.~), (^.))
 import Colog (pattern D, pattern E, pattern I, LogAction (..), Severity (..), pattern W, log)
 
 import Data.Aeson as Json (FromJSON (parseJSON), ToJSON (toJSON))
+
+import Data.Text (splitOn, strip)
 
 import Database.PostgreSQL.Simple.FromField as Sql (FromField (fromField))
 import Database.PostgreSQL.Simple.FromRow as Sql (FromRow (fromRow), field)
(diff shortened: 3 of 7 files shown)


src/IW/Db/Issue.hs
```diff
@@ -4,48 +4,49 @@
...
-getIssuesByLabel :: (WithDb env m) => Text -> m [WithId Issue]
-getIssuesByLabel = query [sql|
+-- | Returns all issues with at least one label in the given @SqlArray@.
+getIssuesByLabels :: (WithDb env m) => SqlArray Label -> m [WithId Issue]
+getIssuesByLabels = query [sql|
     SELECT id, repo_owner, repo_name, number, title, body, labels
-    FROM issues 
-    WHERE ? = ANY (labels)
+    FROM issues
+    WHERE ? && labels
```
chshersh: Can you write an explanation on how this works or provide a link to the documentation? :slightly_smiling_face: I'm just curious

rashadg1030: @chshersh I found this right here: 
https://stackoverflow.com/questions/21742929/postgresql-check-if-array-contains-any-element-from-left-hand-array
If ANY element in one array is in the other it returns true.

src/IW/Server/Issue.hs
```diff
@@ -27,11 +27,11 @@ issueServer = IssueSite
...
-issuesHandler 
+issuesHandler
     :: ( WithDb env m
        )
     => Maybe Text
     -> m [WithId Issue]
 issuesHandler maybeLabel = case maybeLabel of
     Nothing    -> getIssues
-    Just label -> getIssuesByLabel label 
+    Just label -> undefined
```
vrom911: Sorry, is `getIssuesByLabel` not suiting here anymore?

rashadg1030: @vrom911 No it is I just have to modify the function because it’s a list and not just a single label anymore. I’m doing that now.

rashadg1030:
I added a function to the Prelude called splitAndStrip. I think it's general enough to go there, and it can be used in other parts of the code base. I use the same function in branch #119.

src/IW/Db/Issue.hs
```diff
@@ -4,48 +4,49 @@
...
+getIssuesByLabels :: (WithDb env m) => [Label] -> m [WithId Issue]
+getIssuesByLabels = query [sql|
     SELECT id, repo_owner, repo_name, number, title, body, labels
-    FROM issues 
-    WHERE ? = ANY (labels)
+    FROM issues
+    WHERE ? && labels
     LIMIT 100
-|] . Only
+|] . Only . SqlArray
```
chshersh: This is a good function to be used with named parameters later. So you can get rid of `Only` and just use:

```haskell
queryNamed [sql|
...
|] [ "labels" =? SqlArray labels ]
```

src/IW/Db/Issue.hs
```diff
@@ -4,48 +4,49 @@
...
 getIssues = queryRaw [sql|
     SELECT id, repo_owner, repo_name, number, title, body, labels
     FROM issues
     LIMIT 100
 |]
 
--- | Returns a list of issues filtered by label name
-getIssuesByLabel :: (WithDb env m) => Text -> m [WithId Issue]
-getIssuesByLabel = query [sql|
+-- | Returns all issues with at least one label in the given list.
```
chshersh: Documentation is clear! Though, it's an open question, how do we want to model filtering. For example, when a user enters labels `difficulty:easy` and `difficulty:medium` does he want to have issues that have both labels `easy` and `medium` or any of the labels? In this case, probably the first. But what if here enters `difficulty:easy` and `refactoring`? Maybe now he wants to have these two labels. We can't read the user's mind so we either will stick to one default or support both strategies. Not now, just something to think about.

src/IW/Server/Issue.hs
```diff
@@ -7,16 +7,16 @@ module IW.Server.Issue
...
-import IW.Db (WithDb, getIssues, getIssuesByLabel)
+import IW.Db (WithDb, getIssues, getIssuesByLabels)
 import IW.Server.Types (AppServer, ToApi)
 
 
 newtype IssueSite route = IssueSite
     { issuesRoute :: route
         :- "issues"
-        :> QueryParam "label" Text
+        :> QueryParam "labels" Text
```
chshersh: If you want to pass the list of labels, I propose to replace `QueryParam` with `ReqBody` for simplicity. So the list of labels will be given as a JSON request body. For now, we have the only list of labels, but down the road, we might have a complicated filtering structure. So it's easier just to pass it a body.

```suggestion
        :> ReqBody '[JSON] [Label]
```

test/Test/Db.hs
```diff
@@ -52,26 +57,34 @@ upsertReposSpec env = describe "upsertRepos" $ do
...
-        env & reposUnaffected getRepos `equals` True 
+    it "should should not affect the repos table" $
+        env & reposUnaffected getRepos `equals` True
+
+getReposByCategoriesSpec :: AppEnv -> Spec
+getReposByCategoriesSpec env = describe "getReposByCategories" $ do
+    it "should return a list of repos with length 1" $
+        env & rowsReturned (getReposByCategories [Category "FFI"]) `equals` 1
+    it "should not affect the repos table" $
+        env & reposUnaffected (getReposByCategories []) `equals` True
```
chshersh: I think we don't have to be too paranoid about own code. We use `query` with SELECT there, it won't affect any rows.

test/Test/Db.hs
```diff
@@ -52,26 +57,34 @@ upsertReposSpec env = describe "upsertRepos" $ do
...
+    it "should return a list of repos with length 1" $
         env & reposRowCount `equals` 1
-        env & reposUnaffected getRepos `equals` True 
+    it "should should not affect the repos table" $
+        env & reposUnaffected getRepos `equals` True
+
+getReposByCategoriesSpec :: AppEnv -> Spec
+getReposByCategoriesSpec env = describe "getReposByCategories" $ do
+    it "should return a list of repos with length 1" $
+        env & rowsReturned (getReposByCategories [Category "FFI"]) `equals` 1
```
chshersh: It seems that testing on comparison with the specific list of issues is more robust than just checking the length of the list
```suggestion
        env & getReposByCategories [Category "FFI"]) `equals` [testIssue]
```

src/IW/Db/Issue.hs
rashadg1030: @chshersh Good question :thinking:

src/IW/Server/Issue.hs
rashadg1030: @chshersh I see :bulb:

test/Test/Db.hs
rashadg1030: @chshersh Got it :+1:

test/Test/Db.hs
rashadg1030: @chshersh Got it :+1:

src/IW/Db/Issue.hs
rashadg1030: @chshersh I was going to introduce this now, but I think I'm going to open up another pull request and introduce posgresql-simple-named there and just convert some of the functions where it makes sense.

chshersh: @rashadg1030 Yep, it's a good idea to do this in a separate PR 👍

src/Prelude.hs
```diff
@@ -42,3 +45,12 @@ import qualified Colog (Message, WithLog)
...
 -- | 'Colog.WithLog' alias specialized to 'Message' data type.
 type WithLog env m = Colog.WithLog env Colog.Message m
+
+{- | This function takes a delimeter and a delimeter seperated value,
+and returns a list of @Text@ values stripped of excess whitespace.
+Note that it returns an empty list when an empty delimeter seperated value is
+passed in. This prevents the value @[""]@ from being returned.
+-}
+splitAndStrip :: Text -> Text -> [Text]
+splitAndStrip _ ""       = []
```
chshersh: This function is not used in this PR. But it's used in another PR with `MonadCabal` (if I'm not wrong). Need to remember to not have duplication after resolving conflicts.

rashadg1030: That’s correct.

rashadg1030: @chshersh I fixed this.

PR #123, opened 2019-07-13
[#104] Add functionality for associating repos with their categories

### LINKED ISSUE #104: Add functionality for associating Repo categories with the Repo
We currently fetch the repo data from GitHub and parse the repo categories from the cabal file separately. We must now combine them to produce a whole repo that can be stored in the DB.

Blocked by #119

rashadg1030:
Resolves #104

### DIFF
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -96,8 +96,10 @@ library
 
                          -- Sync Worker
                          IW.Sync.Search
+                         IW.Sync.Update
 
   build-depends:       aeson >= 1.4
+                     , async ^>= 2.2.2
                      , bytestring ^>= 0.10
                      , Cabal ^>= 2.4.1.0
                      , case-insensitive ^>= 1.2
@@ -109,6 +111,7 @@ library
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
+                     , lifted-async ^>= 0.10.0.4
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
                      , random ^>= 1.1
@@ -118,6 +121,8 @@ library
                      , text
                      , time >= 1.8 && < 1.10
                      , tomland ^>= 1.0.0
+                     , unliftio ^>= 0.2.12
+                     , unliftio-core ^>= 0.1.2.0
                      , unordered-containers
                      , vector ^>= 0.12.0.3
                      , warp ^>= 3.2
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -111,7 +111,6 @@ library
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
-                     , lifted-async ^>= 0.10.0.4
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
                      , random ^>= 1.1
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -56,6 +56,7 @@ throwError = E.throwError . AppError (toSourcePosition callStack)
 -- | Specialized version of 'E.catchError'.
 catchError :: WithError m => m a -> (AppErrorType -> m a) -> m a
 catchError action handler = action `E.catchError` (handler . appErrorType)
+{-# INLINE catchError #-}
 
 newtype SourcePosition = SourcePosition Text
     deriving newtype (Show, Eq)
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -7,6 +7,7 @@ module IW.App.Error
        , WithError
        , githubErrToAppErr
        , throwError
+       , catchError
        , toHttpError
 
          -- * Error checks
@@ -22,6 +23,7 @@ module IW.App.Error
        , missingHeader
        , headerDecodeError
        , dbError
+       , urlDownloadFailedError
 
          -- * Error throwing helpers
        , throwOnNothing
@@ -38,7 +40,7 @@ import Servant.Server (err401, err404, err417, err500, errBody)
 
 import IW.Core.Url (Url (..))
 
-import qualified Control.Monad.Except as E (throwError)
+import qualified Control.Monad.Except as E (throwError, catchError)
 import qualified GitHub
 import qualified Servant.Server as Servant (ServerError)
 
@@ -46,10 +48,14 @@ import qualified Servant.Server as Servant (ServerError)
 -- | Type alias for errors.
 type WithError m = (MonadError AppError m, HasCallStack)
 
--- | Specialized version of 'E.throwError'
+-- | Specialized version of 'E.throwError'.
 throwError :: WithError m => AppErrorType -> m a
 throwError = E.throwError . AppError (toSourcePosition callStack)
 {-# INLINE throwError #-}
+
+-- | Specialized version of 'E.catchError'.
+catchError :: WithError m => m a -> (AppErrorType -> m a) -> m a
+catchError action handler = action `E.catchError` (handler . appErrorType)
 
 newtype SourcePosition = SourcePosition Text
     deriving newtype (Show, Eq)
@@ -191,6 +197,9 @@ headerDecodeError = InternalError . HeaderDecodeError
 
 dbError :: Text -> AppErrorType
 dbError = InternalError . DbError
+
+urlDownloadFailedError :: Url -> AppErrorType
+urlDownloadFailedError = UrlDownloadFailed
 
 ----------------------------------------------------------------------------
 -- Helpers
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -36,6 +36,8 @@ import GHC.Stack (SrcLoc (SrcLoc, srcLocModule, srcLocStartLine))
 import Network.HTTP.Types.Header (HeaderName)
 import Servant.Server (err401, err404, err417, err500, errBody)
 
+import IW.Core.Url (Url (..))
+
 import qualified Control.Monad.Except as E (throwError)
 import qualified GitHub
 import qualified Servant.Server as Servant (ServerError)
@@ -81,9 +83,10 @@ data AppError = AppError
     } deriving (Show, Eq)
 
 -- | App errors type.
-data AppErrorType 
+data AppErrorType
     = InternalError IError
     | GitHubError GError
+    | UrlDownloadFailed Url
     deriving (Show, Eq)
 
 {- | The internal errors that can be thrown. These errors are meant to be
@@ -113,9 +116,9 @@ data IError
     | DbError Text
     deriving (Show, Eq)
 
-{- | Errors from the @github@ library search functions that can be thrown. 
+{- | Errors from the @github@ library search functions that can be thrown.
 -}
-data GError 
+data GError
     {- | A HTTP error occurred. The actual caught error is included. -}
     = HTTPError Text
     {- | An error in the parser itself. -}
@@ -129,11 +132,11 @@ data GError
 -- | Map the @github@ library's @Error@ type into AppErrorType.
 githubErrToAppErr :: GitHub.Error -> AppErrorType
 githubErrToAppErr = \case
-    GitHub.HTTPError httpException -> GitHubError $ HTTPError $ show httpException 
-    GitHub.ParseError text         -> GitHubError $ ParseError text 
+    GitHub.HTTPError httpException -> GitHubError $ HTTPError $ show httpException
+    GitHub.ParseError text         -> GitHubError $ ParseError text
     GitHub.JsonError text          -> GitHubError $ JsonError text
     GitHub.UserError text          -> GitHubError $ UserError text
- 
+
 -- | Map 'AppError' into a HTTP error code.
 toHttpError :: AppError -> Servant.ServerError
 toHttpError (AppError _callStack errorType) = case errorType of
@@ -146,6 +149,7 @@ toHttpError (AppError _callStack errorType) = case errorType of
         HeaderDecodeError name -> err401 { errBody = encodeUtf8 $ "Unable to decode header: " <> name }
         DbError e              -> err500 { errBody = encodeUtf8 e }
     GitHubError err -> err500 { errBody = show err }
+    UrlDownloadFailed url -> err500 { errBody = encodeUtf8 $ "Couldn't download file from " <> unUrl url }
 
 ----------------------------------------------------------------------------
 -- Error checks
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -2,7 +2,7 @@
 
 module IW.App.Error
        ( AppError (..)
-       , AppErrorType
+       , AppErrorType (..)
        , AppException (..)
        , WithError
        , githubErrToAppErr
--- a/src/IW/Db/Repo.hs
+++ b/src/IW/Db/Repo.hs
@@ -6,12 +6,14 @@ module IW.Db.Repo
        ( getRepos
        , getReposByCategories
        , upsertRepos
+       , updateRepoCategories
        ) where
 
 import IW.Core.Repo (Repo (..), Category (..))
 import IW.Core.SqlArray (SqlArray (..))
 import IW.Core.WithId (WithId)
-import IW.Db.Functions (WithDb, executeMany, query, queryRaw)
+import IW.Db.Functions (WithDb, executeMany, execute, query, queryRaw)
+import IW.Effects.Cabal (MonadCabal (..))
 
 
 -- | Returns all repos in the database.
@@ -43,3 +45,14 @@ upsertRepos = executeMany [sql|
         descr = EXCLUDED.descr
       , categories = EXCLUDED.categories;
 |]
+
+-- | Update a repo's categories field.
+updateRepoCategories :: (MonadCabal m, WithDb env m) => Repo -> m ()
+updateRepoCategories Repo{..} = do
+    categories <- getCabalCategories repoOwner repoName
+    execute [sql|
+        UPDATE repos
+        SET categories = ?
+        WHERE
+            (owner, name) == (?,?)
+    |] (SqlArray categories, repoOwner, repoName)
--- a/src/IW/Db/Repo.hs
+++ b/src/IW/Db/Repo.hs
@@ -53,6 +53,6 @@ updateRepoCategories Repo{..} = do
     execute [sql|
         UPDATE repos
         SET categories = ?
-        WHERE
-            (owner, name) = (?, ?)
+        WHERE owner = ?
+        AND   name  = ?
     |] (SqlArray categories, repoOwner, repoName)
--- a/src/IW/Db/Repo.hs
+++ b/src/IW/Db/Repo.hs
@@ -16,7 +16,7 @@ import IW.Db.Functions (WithDb, executeMany, execute, query, queryRaw)
 
 
 -- | Returns all repos in the database.
-getRepos :: (WithDb env m) => m [WithId Repo]
+getRepos :: WithDb env m => m [WithId Repo]
 getRepos = queryRaw [sql|
     SELECT id, owner, name, descr, categories
     FROM repos
@@ -24,7 +24,7 @@ getRepos = queryRaw [sql|
 |]
 
 -- | Returns all repos with at least one category in the given list.
-getReposByCategories :: (WithDb env m) => [Category] -> m [WithId Repo]
+getReposByCategories :: WithDb env m => [Category] -> m [WithId Repo]
 getReposByCategories = query [sql|
     SELECT id, owner, name, descr, categories
     FROM repos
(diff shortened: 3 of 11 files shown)


rashadg1030:
I wanted to use `mapConcurrently_ :: (a -> IO b) -> t a -> IO ()` but its not `IO` actions so I have to use `mapConcurrently_ :: (Traversable t, MonadBaseControl IO m) => (a -> m b) -> t a -> m ()`. In order to use the second one I have to implement a few instances for `MonadBaseControl IO m` I believe and I was wondering if it is worth it?

rashadg1030:
```
fetchAndUpsertRepos
    ::
    ( MonadCabal m
    , WithDb env m
    , WithLog env m
    , WithError m
    )
    => m ()
fetchAndUpsertRepos = do
    gitHubRepos <- fetchAllHaskellRepos
    let repos = fromGitHubRepo <$> gitHubRepos
    upsertRepos repos
    liftIO $ mapConcurrently_ updateRepoCategories repos
```

chshersh:
@rashadg1030 Okay, let's fix problems one by one :slightly_smiling_face: 

> `mapConcurrently_ :: (a -> IO b) -> t a -> IO ()`

It is a known problem in Haskell and effect systems. And it's a difficult problem. Fortunately, in our case, it can be solved easily. When a function returns `IO`, then it's easy, you just add `liftIO :: IO a -> m a` to the result. However, if the function is expecting `IO` as an argument and you want to pass your custom monadic action, you need to have a function `toIO :: m a -> IO a` which is not possible to write in general. `MonadBaseControl` is one of the first attempts to solve this problem, and the solution is poor.. In our production using `MonadBaseControl` led to a bug which was extremely difficult to find. In our case, we need to use `MonadUnliftIO` constraint. The following blog post describes `MonadUnliftIO` nicely:

* https://fbrs.io/unliftio/

So, you need to derive `MonadUnliftIO` for `App` and use `mapConcurrently_` from `unliftio` package:

* https://hackage.haskell.org/package/unliftio-0.2.12/docs/UnliftIO-Async.html#v:mapConcurrently_

> I have to specify WithLog env m as a constraint in WithCabal eventhough:

This is a question about how typeclass constraints are inferred. The short answer here: if you add only `MonadDownload` to constraints then the only polymorphic function you can use is `downloadFile`. `MonadDownload` has only a single superclass: `Monad`. So you can use `do`-notation, `pure` and `downloadFile`. `WithLog` constraint is a constraint of `downloadFileImpl`, but you don't use `downloadFileImpl`. Constraints are automatically added only if they are superclasses of other constraints. For example, if you have `Ord a` you don't need to write `Eq a` in your constraints. Same thing here.

rashadg1030:
@chshersh Thank you!

rashadg1030:
@chshersh Now that I have implemented a `fetchAndUpsertRepos` function, I tried testing it and got this:
```> env <- mkGhciEnv
> runAppLogIO env $ fetchAndUpsertRepos
[Info]    [14-07-2019 22:31:24.247] [IW.Sync.Search.liftGitHubSearchToApp#81] [ThreadId 219] Fetching total of 90393 Repos...
*** Exception: stack overflow
```
Stack overflow...I'm wondering if it's because I don't have an explicit instance of `MonadUnliftIO`?
```
instance MonadUnliftIO App where
```
I jsut have this, is it necessary to implement it and use `INLINE` pragmas?

chshersh:
When you write instance like this:

```haskell
instance MyClass MyType where
```

then the methods are defined like this:

```
method1 = undefined
method2 = undefined
```

Methods will be implemented properly **only** if they have `default` signatures. In your case instead of writing `instance MonadUnliftIO App` you need to add `MonadUnliftIO` to **deriving**

rashadg1030:
@chshersh okay I get it now. Makes sense thank you.

chshersh:
@rashadg1030 

> Stack overflow...I'm wondering if it's because I don't have an explicit instance of MonadUnliftIO?

Actually, in your case, you see `stack overflow` because `MonadUnliftIO` has two methods, and they have mutually recursive definitions. So the program hangs then these two functions call each other in the infinite loop. If some functions don't have default definitions, GHC gives you a warning. But you haven't seen a warning because the definitions are there like with the `Eq` typeclass. I can see how this error can be confusing.

rashadg1030:
I'm debugging this function:
```

fetchAndUpsertRepos
    ::
    ( MonadCabal m
    , MonadUnliftIO m
    , WithDb env m
    , WithLog env m
    , WithError m
    )
    => m ()
fetchAndUpsertRepos = do
    gitHubRepos <- fetchAllHaskellRepos
    let repos = fromGitHubRepo <$> gitHubRepos
    upsertRepos repos
    mapConcurrently_ updateRepoCategories repos
```
I don't get a stack overflow anymore which is great, but I'm gettiing this error after trying to update 30 repos:

```
*** Exception: SqlError {sqlState = "42883", sqlExecStatus = FatalError, sqlErrorMsg = "operator does not exist: text == text", sqlErrorDetail = "", sqlErrorHint = "No operator matches the given name and argument type(s). You might need to add explicit type casts."}
```

I believe the error is coming from this function:

-- | Update a repo's categories field.
updateRepoCategories :: (MonadCabal m, WithDb env m) => Repo -> m ()
updateRepoCategories Repo{..} = do
    categories <- getCabalCategories repoOwner repoName
    execute [sql|
        UPDATE repos
        SET categories = (? :: TEXT[])
        WHERE
            owner == (? :: TEXT) && name == (? :: TEXT)
    |] (SqlArray categories, repoOwner, repoName)
```
I'm adding type annotations because that's what the error reccomends. Without the type annotations the error is pretty much the same except:
```
*** Exception: SqlError {sqlState = "42883", sqlExecStatus = FatalError, sqlErrorMsg = "operator does not exist: text == unknown", sqlErrorDetail = "", sqlErrorHint = "No operator matches the given name and argument type(s). You might need to add explicit type casts."}
```
It goes from `text == text` to `text == unknown` without type annotations.

**Comment by rashadg1030** [comment]
Also, it seems like we can only get 30 repos a request. I will look more into this, but it seems like we need to change the query string we're using for the `github` API to get more results. 
```
Fetching total of 90401 Repos...
```
It should return 90401 repos.

**Comment by chshersh** [comment]
@rashadg1030 SQL is weird and not consistent with normal programming languages. The problem is in `==` operator. In SQL use sing `=` to check for equality.

**Comment by rashadg1030** [comment]
@chshersh Wow, thank you. It works great now. Honestly I had no idea `=` was used for equality. I thought it was strictly used for assignment :sweat_smile:

**Comment by rashadg1030** [comment]
https://stackoverflow.com/questions/30656761/github-search-api-only-return-30-results
The above link provides some insight into how to get more than results per request. 

I would also like to mention that I removed a section of `downloadFile` where an error is thrown so the program keeps continuing even if it can't download a file.

**Comment by chshersh** [comment]
@rashadg1030 Yeah, I kinda expected that GitHub implements pagination for their own API. Otherwise, it would die trying to serve all requests 😂 We need to implement a separate function to solve this problem and I propose to implement this function in a separate PR because it's going to be complicated:

1. The function should download repos page by page.
2. The function should be configurable, how many results per page and how many pages should work (so we can test it locally without downloading all 90K repos).
3. It should use some ordering so it will be more or less deterministic in terms of results.
4. Since we process all repos in chunks, you need to think how to combine all chunks into a single list of values. Maybe at some point we will need to use some streaming library... But let's worry about optimizations later. First, we need to make it work.

**Comment by rashadg1030** [comment]
I think it's best if we merge this, and then I'll open up a new issue and PR for handling GitHub API pagination.

> @rashadg1030 Yeah, I kinda expected that GitHub implements pagination for their own API. Otherwise, it would die trying to serve all requests  We need to implement a separate function to solve this problem and I propose to implement this function in a separate PR because it's going to be complicated:
> 
> 1. The function should download repos page by page.
> 2. The function should be configurable, how many results per page and how many pages should work (so we can test it locally without downloading all 90K repos).
> 3. It should use some ordering so it will be more or less deterministic in terms of results.
> 4. Since we process all repos in chunks, you need to think how to combine all chunks into a single list of values. Maybe at some point we will need to use some streaming library... But let's worry about optimizations later. First, we need to make it work.

File: issue-wanted.cabal [hunk 1]
```diff
@@ -109,6 +111,7 @@ library
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
+                     , lifted-async ^>= 0.10.0.4
```
chshersh: This library can be removed from dependencies since we are using `unliftio` now

src/IW/Db/Repo.hs
```diff
@@ -43,3 +45,14 @@ upsertRepos = executeMany [sql|
...
+
+-- | Update a repo's categories field.
+updateRepoCategories :: (MonadCabal m, WithDb env m) => Repo -> m ()
+updateRepoCategories Repo{..} = do
+    categories <- getCabalCategories repoOwner repoName
+    execute [sql|
+        UPDATE repos
+        SET categories = ?
+        WHERE
+            (owner, name) = (?, ?)
```
chshersh: Instead of `(x, y) = (?, ?)` it's better to use `AND` operator. It's more idiomatic SQL than using tuples in SQL:

```haskell
WHERE owner = ?
  AND name = ?
```

src/IW/Effects/Cabal.hs
```diff
@@ -63,13 +62,7 @@ repoCabalUrl (RepoOwner repoOwner) (RepoName repoName) = Url $
...
   where
     splitCategories :: GenericPackageDescription -> [Text]
-    splitCategories = splitOnNonEmptyText "," . toText . category . packageDescription
-
-    -- | A variation of @splitOn@ that returns @[]@ instead of @[""]@
-    -- if the text to be split is empty.
-    splitOnNonEmptyText :: Text -> Text -> [Text]
-    splitOnNonEmptyText _ ""       = []
-    splitOnNonEmptyText delim text = splitOn delim text
+    splitCategories = splitAndStrip "," . toText . category . packageDescription
```
chshersh: If I recall correctly, `splitAntStrip` function is now in `Prelude`. I propose to move it in this module. It's not going to be used in many modules (probably it's going to be used only here), so no need to move into `Prelude`. Candidates for adding to `Prelude` are functions or types that used in at least half of the project and unambiguous.

src/IW/Db/Repo.hs
```diff
@@ -43,3 +45,14 @@ upsertRepos = executeMany [sql|
         descr = EXCLUDED.descr
       , categories = EXCLUDED.categories;
 |]
+
+-- | Update a repo's categories field.
+updateRepoCategories :: (MonadCabal m, WithDb env m) => Repo -> m ()
+updateRepoCategories Repo{..} = do
+    categories <- getCabalCategories repoOwner repoName
+    execute [sql|
```
chshersh: A good practice that we learned during development is that it's usually a good idea to put SQL queries in separate functions and not interleave them with other effects. That leads to simpler code and helps with testing. For example, if you want to test `updateRepoCategories` function, you need to pass the whole repo, and the function will download something. So for the tests for this function to work, you actually need to give valid testing data connected to the real world. Also, there is no much benefit in testing `getCabalCategories + SQL` query because we already have tests for `getCabalCategories` so need to test only SQL queries.

Since we're writing raw SQL in Haskell code, we need to test every our query (that it doesn't contain syntax errors, that it works as we expect). So we can make our lives easier by decomposing all SQL queries into separate functions. The most difficult functions to test in our codebase are the functions that do multiple actions of different effects inside them. Haskell allows to decompose problems into smaller pieces, test them independently and combine with confidence. We need to write only a couple of tests for `getCabalCategories`, and we will be sure that the function works. Same for SQL queries. And then we don't need to write tests for functions that just compose multiple tested functions (not always true if logic is not trivial, but this is what we're trying to achieve ultimately).

src/IW/Sync/Update.hs
```diff
@@ -0,0 +1,25 @@
+module IW.Sync.Update where
```
chshersh: Let's add some docs to this module and explicit export list

src/IW/Sync/Update.hs
```diff
@@ -0,0 +1,25 @@
...
+
+import Control.Monad.IO.Unlift (MonadUnliftIO)
+import UnliftIO.Async (mapConcurrently_)
+
+import IW.App (WithError)
+import IW.Db
+import IW.Effects.Cabal (MonadCabal (..))
+import IW.Sync.Search
```
chshersh: Some imports here are implicit

src/IW/Sync/Update.hs
```diff
@@ -0,0 +1,25 @@
...
+
+
+fetchAndUpsertRepos
+    ::
+    ( MonadCabal m
```
chshersh: According to our style guide, we shouldn't have hanging `::`. This code should be formatted like this:

```haskell
fetchAndUpsertRepos
    :: ( MonadCabal m
       , MonadUnliftIO m
       ...
       )
    => m ()
```

src/IW/Sync/Update.hs
```diff
@@ -0,0 +1,25 @@
...
+    , MonadUnliftIO m
+    , WithDb env m
+    , WithLog env m
+    , WithError m
+    )
+    => m ()
+fetchAndUpsertRepos = do
+    gitHubRepos <- fetchAllHaskellRepos
+    let repos = fromGitHubRepo <$> gitHubRepos
```
chshersh: I usually prefer to use `map` when working with ordinary lists. Sometimes using monomorphic functions over polymorphic ones have readability benefits 🙂

src/IW/Effects/Cabal.hs
rashadg1030: @chshersh I agree

src/IW/Sync/Update.hs
rashadg1030: @chshersh Right!

src/IW/Db/Repo.hs
rashadg1030: @chshersh Understood. Makes sense :bulb:

src/IW/Sync/Update.hs
rashadg1030: @chshersh I will fix this right now.

src/IW/Sync/Update.hs
rashadg1030: @chshersh That's true :+1:

rashadg1030:
@chshersh This PR is good to go.

src/IW/Db/Repo.hs
```diff
@@ -43,3 +44,18 @@ upsertRepos = executeMany [sql|
         descr = EXCLUDED.descr
       , categories = EXCLUDED.categories;
 |]
+
+-- | Update a repo's categories field.
+updateRepoCategories
+    :: ( WithDb env m
+       )
```
chshersh: When there is only a single constraint, no need to spawn multiple redundant lines. This can be just:

```haskell
    :: WithDb env m
    =>

File: src/IW/Effects/Download.hs [hunk 10]
```diff
@@ -41,4 +41,4 @@ downloadFileImpl Url{..} = do
             pure $ toStrict body
         _   -> do
             log W $ "Couldn't download file from " <> unUrl
-            throwError notFound
+            pure ""
```
chshersh: This is actually a bit unpleasant moment. I think that if a file is failed to download, this should be an error, not just empty string with logging. And then later in the `getCabalCategories` you can handle the error appropriately using `catchError` function. I would add one more constructor `UrlDownloadFailed Url` to the `AppErrorType` data type to have a descriptive name of this error instead of generic `notFound`. But the general idea is that the `MonadDownload` effect can be used not only by `MonadCabal` and it shouldn't know about the implementation details of classes that are going to use it. Not being able to download a fiel is a valid error.

rashadg1030: @chshersh got it. I forgot to ask about this.

rashadg1030: @chshersh 
```
getCabalCategoriesImpl
    :: forall env m.
       WithCabal env m
    => RepoOwner
    -> RepoName
    -> m [Category]
getCabalCategoriesImpl repoOwner repoName = do
    cabalFile <- downloadFile cabalUrl `catchError` urlDownloadFailedHandler
    case parseGenericPackageDescriptionMaybe cabalFile of
        Nothing -> do
            log W $ "Couldn't parse file downloaded from " <> unUrl cabalUrl
            pure []
        Just genPkgDescr -> do
            log I $ "Successfully parsed file downloaded from " <> unUrl cabalUrl
            pure $ categoryNames genPkgDescr
  where
    cabalUrl :: Url
    cabalUrl = repoCabalUrl repoOwner repoName

    urlDownloadFailedHandler :: AppErrorType -> m ByteString
    urlDownloadFailedHandler _ = pure ""
```
Is this fine for handling the error? Or is there a better way? I don't like the fact that the handler ignores the `AppErrorType`, but at the same time pattern matching on each error seems like excessive when it doesn't matter. Maybe the `catchError` should be later in the `do` block? After the part where the `cabalFile` is parsed.

rashadg1030: This works as well and looks better imo:
```
getCabalCategoriesImpl
    :: forall env m.
       WithCabal env m
    => RepoOwner
    -> RepoName
    -> m [Category]
getCabalCategoriesImpl repoOwner repoName = do
    { cabalFile <- downloadFile cabalUrl;
      case parseGenericPackageDescriptionMaybe cabalFile of
          Nothing -> do
              log W $ "Couldn't parse file downloaded from " <> unUrl cabalUrl
              pure []
          Just genPkgDescr -> do
              log I $ "Successfully parsed file downloaded from " <> unUrl cabalUrl
              pure $ categoryNames genPkgDescr
    } `catchError` handler
  where
    cabalUrl :: Url
    cabalUrl = repoCabalUrl repoOwner repoName

    handler :: AppErrorType -> m [Category]
    handler _ = pure []
```

chshersh: @rashadg1030 `do` with curly braces is weird... Let's leave this syntax to GHC developers 🙂 

I indeed like the second option better because returning `empty` list rather than trying to parse empty text is a saner approach. However, your `catchError` is attached to the whole function body where in reality we are interested in catching error only for `downloadFile` function. It doesn't affect behavior in this case (though it can in more general case) but it makes code more readable.

If I had to write this function, I would implement it like this. First, introduce `downloadFileMaybe` wrapper around `downloadFile`:

```haskell
downloadFileMaybe :: (MonadDownload m, WithError m) => Url -> m (Maybe ByteString)
downloadFileMaybe url = downloadFile url `catchError` \case
    UrlDownloadFailed _ -> pure Nothing
    err -> throwError err
```

And then you can use this function:

```haskell
getCabalCategoriesImpl repoOwner repoName = do
    maybeCabalFile <- downloadFileMaybe cabalUrl
    case maybeCabalFile >>= parseGenericPackageDescriptionMaybe of
        ...
```

(didn't compile, so there might be some mistakes, but I hope that idea is more or less clear)

rashadg1030: @chshersh This is clear👍🏽 Thank you.

rashadg1030: @chshersh It won't compile becasue the type of `catchError` doesn't allow it. 
```
-- | Specialized version of 'E.catchError'.
catchError :: WithError m => m a -> (AppErrorType -> m a) -> m a
catchError action handler = action `E.catchError` (handler . appErrorType)
{-# INLINE catchError #-}

downloadFileMaybe :: MonadDownload m => Url -> m (Maybe ByteString)
downloadFileMaybe url = downloadFile url `catchError` \case
    UrlDownloadFailed _ -> pure Nothing
    err -> throwError err
```
`downloadFile` would have to return a `m (Maybe ByteString)`, but it just returns a `m ByteString`. Should the `downloadFileImpl` be modified to return the correct type?

rashadg1030: I also tried to use something like this, but didn't work out

```
-- | Specialized version of 'E.catchError'.
catchErrorMaybe :: WithError m => m a -> (AppErrorType -> m (Maybe a)) -> m (Maybe a)
catchErrorMaybe action handler = action `E.catchError` (handler . appErrorType)
{-# INLINE catchErrorMaybe #-}
```

chshersh: @rashadg1030 This is a common case. Try this code:

```haskell
(Just <$> downloadFile url) `catchError` \case
```

Functors for the win

rashadg1030: @chshersh Bingo! Thank you.

src/IW/App/Error.hs
```diff
@@ -36,19 +38,26 @@ import GHC.Stack (SrcLoc (SrcLoc, srcLocModule, srcLocStartLine))
...
--- | Specialized version of 'E.throwError'
+-- | Specialized version of 'E.throwError'.
 throwError :: WithError m => AppErrorType -> m a
 throwError = E.throwError . AppError (toSourcePosition callStack)
 {-# INLINE throwError #-}
 
+-- | Specialized version of 'E.catchError'.
+catchError :: WithError m => m a -> (AppErrorType -> m a) -> m a
+catchError action handler = action `E.catchError` (handler . appErrorType)
+{-# INLINE catchError #-}
```
chshersh: This is a useful function! Nice addition to the interface :+1:

PR #127, opened 2019-07-15
[#126] Implement function for fetching repos with pagination

### LINKED ISSUE #126: Implement function for fetching repos page by page
1. The function should download repos page by page.
2. The function should be configurable, how many results per page and how many pages should work (so we can test it locally without downloading all 90K repos).
3. It should use some ordering so it will be more or less deterministic in terms of results.
4. Since we process all repos in chunks, you need to think how to combine all chunks into a single list of values. Maybe at some point we will need to use some streaming library... But let's worry about optimizations later. First, we need to make it work.

rashadg1030:
Resolves #126

### DIFF
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -111,7 +111,6 @@ library
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
-                     , lifted-async ^>= 0.10.0.4
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
                      , postgresql-simple-named ^>= 0.0.0.0
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -98,6 +98,8 @@ library
                          IW.Sync.Search
                          IW.Sync.Update
 
+                         IW.Time
+
   build-depends:       aeson >= 1.4
                      , async ^>= 2.2.2
                      , bytestring ^>= 0.10
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -111,6 +111,7 @@ library
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
+                     , lifted-async ^>= 0.10.0.4
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
                      , postgresql-simple-named ^>= 0.0.0.0
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -24,7 +24,7 @@ module IW.App.Error
        , headerDecodeError
        , dbError
        , dbNamedError
-       , githubHTTPError
+       , githubHttpError
        , urlDownloadFailedError
 
          -- * Error throwing helpers
@@ -132,7 +132,7 @@ data IError
 -}
 data GError
     {- | A HTTP error occurred. The actual caught error is included. -}
-    = HTTPError Text
+    = HttpError Text
     {- | An error in the parser itself. -}
     | ParseError Text
     {- | The JSON is malformed or unexpected. -}
@@ -144,7 +144,7 @@ data GError
 -- | Map the @github@ library's @Error@ type into AppErrorType.
 githubErrToAppErr :: GitHub.Error -> AppErrorType
 githubErrToAppErr = \case
-    GitHub.HTTPError httpException -> GitHubError $ HTTPError $ show httpException
+    GitHub.HTTPError httpException -> GitHubError $ HttpError $ show httpException
     GitHub.ParseError text         -> GitHubError $ ParseError text
     GitHub.JsonError text          -> GitHubError $ JsonError text
     GitHub.UserError text          -> GitHubError $ UserError text
@@ -208,8 +208,8 @@ dbError = InternalError . DbError
 dbNamedError :: PgNamedError -> AppErrorType
 dbNamedError = InternalError . DbNamedError
 
-githubHTTPError :: Text -> AppErrorType
-githubHTTPError = GitHubError . HTTPError
+githubHttpError :: Text -> AppErrorType
+githubHttpError = GitHubError . HttpError
 
 urlDownloadFailedError :: Url -> AppErrorType
 urlDownloadFailedError = UrlDownloadFailed
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -24,6 +24,7 @@ module IW.App.Error
        , headerDecodeError
        , dbError
        , dbNamedError
+       , githubHTTPError
        , urlDownloadFailedError
 
          -- * Error throwing helpers
@@ -206,6 +207,9 @@ dbError = InternalError . DbError
 
 dbNamedError :: PgNamedError -> AppErrorType
 dbNamedError = InternalError . DbNamedError
+
+githubHTTPError :: Text -> AppErrorType
+githubHTTPError = GitHubError . HTTPError
 
 urlDownloadFailedError :: Url -> AppErrorType
 urlDownloadFailedError = UrlDownloadFailed
--- a/src/IW/Sync/Search.hs
+++ b/src/IW/Sync/Search.hs
@@ -15,7 +15,11 @@ module IW.Sync.Search
        ) where
 
 import GitHub (SearchResult (..), URL (..))
-import GitHub.Endpoints.Search (searchRepos, searchIssues)
+import GitHub.Data.RateLimit (RateLimit (..), Limits)
+import GitHub.Data.Request (query)
+import GitHub.Request (executeRequest')
+import GitHub.Endpoints.Search (searchIssues)
+import GitHub.Endpoints.RateLimit (rateLimit)
 
 import IW.App (WithError)
 import IW.Core.Issue (Issue (..), Label (..))
@@ -34,17 +38,10 @@ fetchAllHaskellRepos
        , WithError m
        , WithLog env m
        )
-    => m [GitHub.Repo]
-fetchAllHaskellRepos = liftGitHubSearchToApp searchHaskellRepos
-
--- | Convert a value of the @GitHub.Repo@ type to a value of our own @Repo@ type.
-fromGitHubRepo :: GitHub.Repo -> Repo
-fromGitHubRepo githubRepo = Repo
-    { repoOwner      = RepoOwner $ GitHub.untagName $ GitHub.simpleOwnerLogin $ GitHub.repoOwner githubRepo
-    , repoName       = RepoName  $ GitHub.untagName $ GitHub.repoName githubRepo
-    , repoDescr      = fromMaybe "" $ GitHub.repoDescription githubRepo
-    , repoCategories = SqlArray []
-    }
+    => Int
+    -> Int
+    -> m [GitHub.Repo]
+fetchAllHaskellRepos perPage page = liftGitHubSearchToApp (searchHaskellRepos perPage page)
 
 -- | Fetch all open issues with Haskell language and the labels passed in to the function.
 fetchHaskellIssuesByLabels
@@ -75,23 +72,57 @@ liftGitHubSearchToApp
        )
     => IO (Either GitHub.Error (SearchResult a))
     -> m [a]
-liftGitHubSearchToApp githubSearch = liftIO githubSearch >>= \case
+liftGitHubSearchToApp searchAction = liftIO searchAction >>= \case
     Left err -> throwError $ githubErrToAppErr err
     Right (SearchResult count vec) -> do
         log Info $ "Fetching total of " <> show count <> " " <> typeName @a <> "s..."
+        searchLimit <- getSearchRateLimit
+        log Info $ "Search rate limit information: " <> show searchLimit
         pure $ V.toList vec
 
--- | Search all repositories built with the Haskell language.
-searchHaskellRepos :: IO (Either GitHub.Error (SearchResult GitHub.Repo))
-searchHaskellRepos = searchRepos "language:haskell"
+getSearchRateLimit
+    :: forall m.
+       ( MonadIO m
+       , WithError m
+       )
+    => m Limits
+getSearchRateLimit = liftIO rateLimit >>= \case
+    Left err -> throwError $ githubErrToAppErr err
+    Right RateLimit{..} -> pure rateLimitSearch
 
--- | Construct a github search query from a list of labels.
-labelsToSearchQuery :: [Label] -> Text
-labelsToSearchQuery = foldMap (\Label{..} -> "label:\"" <> unLabel <> "\" ")
+githubSearch :: FromJSON a => [Text] -> [(ByteString, Maybe ByteString)] -> IO (Either GitHub.Error (SearchResult a))
+githubSearch paths queries = executeRequest' $ query paths queries
+
+{- | Search all repositories built with the Haskell language. Takes an @Int@ representing
+how many results will be returned per page, and another @Int@ that represents the page
+number to be returned.
+-}
+searchHaskellRepos :: Int -> Int -> IO (Either GitHub.Error (SearchResult GitHub.Repo))
+searchHaskellRepos perPage page = githubSearch paths queries
+  where
+    paths   = ["search", "repositories"]
+    queries =
+        [ ("q", Just "language:haskell")
+        , ("per_page", Just $ show perPage)
+        , ("page", Just $ show page)
+        ]
+
+-- | Convert a value of the @GitHub.Repo@ type to a value of our own @Repo@ type.
+fromGitHubRepo :: GitHub.Repo -> Repo
+fromGitHubRepo githubRepo = Repo
+    { repoOwner      = RepoOwner $ GitHub.untagName $ GitHub.simpleOwnerLogin $ GitHub.repoOwner githubRepo
+    , repoName       = RepoName $ GitHub.untagName $ GitHub.repoName githubRepo
+    , repoDescr      = fromMaybe "" $ GitHub.repoDescription githubRepo
+    , repoCategories = SqlArray []
+    }
 
 -- | Search for all open Haskell issues with the corresponding labels.
 searchHaskellIssuesByLabels :: [Label] -> IO (Either GitHub.Error (SearchResult GitHub.Issue))
 searchHaskellIssuesByLabels labels = searchIssues $ "language:haskell is:open " <> labelsToSearchQuery labels
+
+-- | Construct a github search query from a list of labels.
+labelsToSearchQuery :: [Label] -> Text
+labelsToSearchQuery = foldMap (\Label{..} -> "label:\"" <> unLabel <> "\" ")
 
 -- | Convert a value of the @GitHub.Issue@ type to a value of our own @Issue@ type.
 fromGitHubIssue :: GitHub.Issue -> Maybe Issue
--- a/src/IW/Sync/Search.hs
+++ b/src/IW/Sync/Search.hs
@@ -20,8 +20,7 @@ import Data.Time (Day (..))
 import GitHub (SearchResult (..), URL (..), RateLimit (..), Limits, Paths, QueryString, executeRequest', limitsRemaining)
 import GitHub.Endpoints.RateLimit (rateLimit)
 
-import IW.App (WithError)
-import IW.App.Error (githubErrToAppErr, throwError)
+import IW.App (WithError, githubErrToAppErr, githubHTTPError, throwError)
 import IW.Core.Issue (Issue (..), Label (..))
 import IW.Core.Repo (Repo (..), RepoOwner (..), RepoName (..))
 import IW.Core.SqlArray (SqlArray (..))
@@ -43,19 +42,28 @@ liftGithubSearchToApp
        )
     => IO (Either GitHub.Error (SearchResult a))
     -> m [a]
-liftGithubSearchToApp githubSearch' = do
+liftGithubSearchToApp searchAction = do
     searchLimit <- getSearchRateLimit
     if limitsRemaining searchLimit > 0
-        then performSearch
-        else liftGithubSearchToApp githubSearch'
-  where
-    -- | Execute a query against the GitHub Search API.
-    performSearch :: m [a]
-    performSearch = liftIO githubSearch' >>= \case
-        Left err -> throwError $ githubErrToAppErr err
-        Right (SearchResult count vec) -> do
-            log Info $ "Fetched total of " <> show count <> " " <> typeName @a <> "s..."
-            pure $ V.toList vec
+        then execSearch searchAction
+        else throwError $ githubHTTPError "GitHub Search API limit reached."
+
+-- | Execute a query against the GitHub Search API.
+execSearch
+    :: forall a env m.
+       ( MonadIO m
+       , MonadUnliftIO m
+       , WithError m
+       , WithLog env m
+       , Typeable a
+       )
+    => IO (Either GitHub.Error (SearchResult a))
+    -> m [a]
+execSearch searchAction = liftIO searchAction >>= \case
+    Left err -> throwError $ githubErrToAppErr err
+    Right (SearchResult count vec) -> do
+        log Info $ "Fetched total of " <> show count <> " " <> typeName @a <> "s..."
+        pure $ V.toList vec
 
 -- | Function for fetching the current rate limit information for the GitHub Search API.
 getSearchRateLimit :: forall m. (MonadIO m, MonadUnliftIO m, WithError m) => m Limits
--- a/src/IW/Sync/Search.hs
+++ b/src/IW/Sync/Search.hs
@@ -9,6 +9,7 @@ module IW.Sync.Search
        ( fetchHaskellReposByDate
        , fetchAllHaskellIssues
        , fetchHaskellIssuesByLabels
+       , getSearchRateLimit
        , fromGitHubIssue
        , fromGitHubRepo
        , parseUserData
--- a/src/IW/Sync/Search.hs
+++ b/src/IW/Sync/Search.hs
@@ -14,12 +14,13 @@ module IW.Sync.Search
        , parseUserData
        ) where
 
-import UnliftIO (MonadUnliftIO)
 import Data.Time (Day (..), addDays)
 import GitHub (SearchResult (..), URL (..), RateLimit (..), Limits, Paths, QueryString, executeRequest', limitsRemaining)
 import GitHub.Endpoints.RateLimit (rateLimit)
+import UnliftIO (MonadUnliftIO)
+import UnliftIO.Concurrent (threadDelay)
 
-import IW.App (WithError, githubErrToAppErr, githubHTTPError, throwError)
+import IW.App (WithError, githubErrToAppErr, githubHttpError, throwError)
 import IW.Core.Issue (Issue (..), Label (..))
 import IW.Core.Repo (Repo (..), RepoOwner (..), RepoName (..))
 import IW.Core.SqlArray (SqlArray (..))
@@ -99,7 +100,7 @@ githubSearchAll
 githubSearchAll paths properties recent interval acc = do
     let startDay = negate interval `addDays` recent
     -- | Search for first page of query and check the result count to see what to do next.
-    (SearchResult count vec) <- liftGithubSearchToApp $ githubSearch paths properties startDay recent 1
+    (SearchResult count vec) <- liftExecuteSearchToApp $ executeSearch paths properties startDay recent 1
         -- | If count is 0, then all results for the given properties have been obtained.
     if | count == 0   -> do
             log I "All results obtained."
@@ -107,16 +108,17 @@ githubSearchAll paths properties recent interval acc = do
         -- | If count is less than or equal to 1000, then the interval is good and we can get
         -- the rest of the results on pages 2 to 10.
        | count <= 1000 -> do
-            remResults <- foldMapM (githubSearch' paths properties startDay recent) [2..10]
+            remResults <- foldMapM (githubSearch paths properties startDay recent) [2..10]
             githubSearchAll paths properties (pred startDay) interval (acc <> V.toList vec <> remResults)
         -- | Otherwise, call the function with the same arguments but a smaller interval.
        | otherwise    -> githubSearchAll paths properties recent (pred interval) acc
 
-githubSearch'
-    :: forall a m.
+githubSearch
+    :: forall a env m.
        ( MonadIO m
        , MonadUnliftIO m
        , WithError m
+       , WithLog env m
        , FromJSON a
        )
     => Paths -- ^ URL paths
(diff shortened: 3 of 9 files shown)


rashadg1030:
Everything is working well except for one thing. When I try to fetch more than the first 1000 repos, I get this error:
```
{
  "message": "Only the first 1000 search results are available",
  "documentation_url": "https://developer.github.com/v3/search/"
}
```
When you read the docs it only says you are limited to 1000 results at a time, not that you are limited to only the first 1000 results. 
`https://developer.github.com/v3/search/
Other than that, I'm able to get pages 1 - 10 with 100 results each. The code we have right now is able to get categories for a majority of the repos so that's cool. I've also added a way to log rate limiting info like how many requests we have left before the next reset.

chshersh:
@rashadg1030 Hmm, that's weird. I didn't expect GitHub API to have a ceiling on the total amount of search results. This is a show-stopper. Could you please contact GitHub support and ask them for a way to resolve the problem? Mention that you're working on a project as a part of GSoC 2019. Maybe they will issue you a special token 🙂 The response usually takes several days. So it's a good idea to ask on StackOverflow as well. Meanwhile, I found the following workaround (but it could be outdated, probably there is a better solution):

* https://stackoverflow.com/questions/37602893/github-search-limit-results

The idea is to search by `created` date and step through weeks instead of pages. So you need to find the earliest Haskell GitHub project and fetch by 1000 for each week. If the first repo was created in 2013, so we have `52 * (2019 - 2013 + 1) ~ 350` requests to fetch all Haskell repos. The same strategy can be used later for issues. So we are still fit the 5000 requests limit. The only problem if there were more than 1000 Haskell repos or issues created within a week (which I doubt is the case). The SO answer suggests binary search over intervals to narrow the scope even further. But this is too much CS for a simple task. I didn't realise that it would be that hard to fetch data from GitHub...

As a reference, limits for GraphQL GitHub API v4 are different. Unfortunately, GraphQL library in Haskell is not that good... Probably next year somebody will improve GraphQL, and `issue-wanted` can migrate to a newer approach 🙂 

* https://developer.github.com/v4/guides/resource-limitations/

rashadg1030:
> @rashadg1030 Hmm, that's weird. I didn't expect GitHub API to have a ceiling on the total amount of search results. This is a show-stopper. Could you please contact GitHub support and ask them for a way to resolve the problem? Mention that you're working on a project as a part of GSoC 2019. Maybe they will issue you a special token  The response usually takes several days. So it's a good idea to ask on StackOverflow as well. Meanwhile, I found the following workaround (but it could be outdated, probably there is a better solution):
> 
> * https://stackoverflow.com/questions/37602893/github-search-limit-results
> 
> The idea is to search by `created` date and step through weeks instead of pages. So you need to find the earliest Haskell GitHub project and fetch by 1000 for each week. If the first repo was created in 2013, so we have `52 * (2019 - 2013 + 1) ~ 350` requests to fetch all Haskell repos. The same strategy can be used later for issues. So we are still fit the 5000 requests limit. The only problem if there were more than 1000 Haskell repos or issues created within a week (which I doubt is the case). The SO answer suggests binary search over intervals to narrow the scope even further. But this is too much CS for a simple task. I didn't realise that it would be that hard to fetch data from GitHub...
> 
> As a reference, limits for GraphQL GitHub API v4 are different. Unfortunately, GraphQL library in Haskell is not that good... Probably next year somebody will improve GraphQL, and `issue-wanted` can migrate to a newer approach 
> 
> * https://developer.github.com/v4/guides/resource-limitations/

Wow, this is a bummer :slightly_frowning_face: I hate for the project to come to a stop like this. I messaged Github support for a workaround the problem as soon as I found it (yesterday) and still haven't received a response. I will message them again but ask for a token or special access and mention that this is a GSoC project. Hopefully they will understand. This is a shocker because I didn't expect this at all. I'm willing to make some drastic changes to the code to keep the project going if my mentors think it's a good idea. We could use the GraphQL API, but as you said, the Haskell support for it isn't great. Apparently, you can query a GraphQL API over `http`: https://graphql.org/learn/serving-over-http/. Don't know if we can leverage this somehow with the help of `servant-client`. Another far-fetched idea I have is to use the GitLab API.... not sure if it's any different, but GitLab is open-source so I assume we would have full access to their data. Not sure how many Haskell projects are on GitLab though. I've never used it before. 

> The idea is to search by `created` date and step through weeks instead of pages. So you need to find the earliest Haskell GitHub project and fetch by 1000 for each week. If the first repo was created in 2013, so we have `52 * (2019 - 2013 + 1) ~ 350` requests to fetch all Haskell repos. The same strategy can be used later for issues. So we are still fit the 5000 requests limit. The only problem if there were more than 1000 Haskell repos or issues created within a week (which I doubt is the case). The SO answer suggests binary search over intervals to narrow the scope even further. But this is too much CS for a simple task. I didn't realise that it would be that hard to fetch data from GitHub...

I'm willing to try and implement this if there isn't any other way to get around the problem....Let me know what you think. For now, I'm going to message GitHub support again, work on the code a bit more (there's still a few touchups that can be made), and then just wait and hope for the best. We can also still work on the frontend.

chshersh:
@rashadg1030 Well, project didn't stop. We still can implement date-interval pagination as a workaround. It's will be less convenient and less optimal but should work. I saw that other people in other languages went with the same approach due to limitations of the GitHub API. 

Regarding GraphQL: unfortunately, we never used GraphQL, and we don't have the capacity to learn a new way of querying API and understand the library, so we can't mentor in this area and going with GraphQL will be risky. Somebody asked about the state of the GraphQL on Twitter recently:

* https://twitter.com/domenkozar/status/1151060507425673219

Maybe you or somebody else will have time after GSoC to investigate the state of GraphQL in Haskell and replace querying in `issue-wanted` from REST to GraphQL and writing a tutorial-like blog post in the process in a slower pace :slightly_smiling_face: 

> GitLab is open-source so I assume we would have full access to their data. Not sure how many Haskell projects are on GitLab though

I noticed recently that some people from the Haskell community are switching from GitHub to GitLab. However, there are not many projects on GitLab. But the majority of Haskell projects is still on GitHub.

>  For now, I'm going to message GitHub support again, work on the code a bit more

Yeah, no rush here. Last time GitHub support answered me after 3 days or something like that. You can try date-interval strategy. Or focus on writing the blog post while we are waiting for the official response :slightly_smiling_face:

rashadg1030:
@chshersh Okay great. Thanks for that twitter link it was interesting discussion.

rashadg1030:
There is now a function `syncReposByDate` that seems to work pretty nicely. I've tested it, but it can't go back that far before hitting the API rate limit. The next challenge I think will be figuring out how to recover from reaching the API rate limit. The server needs to be able to wait for the rate limit to reset and then query again from where it left off. I will need a timer of some sort I think and poll every minute or so. Also, I'm thinking it would be nice to have a log file for debugging. Write now all the logs are written to the console and I can't see them all because there is so many. Is having a log file a good idea in this situation?

issue-wanted.cabal
```diff
@@ -111,6 +111,7 @@ library
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
+                     , lifted-async ^>= 0.10.0.4
```
chshersh: Is this package needed? I see that we're using `MonadUnliftIO`, not `MonadBaseControl`

src/IW/Sync/Search.hs
```diff
@@ -28,70 +35,117 @@ import qualified Data.Text as T
 import qualified Data.Vector as V
 
 
--- | Fetch all repositories built with the Haskell language.
-fetchAllHaskellRepos
-    :: ( MonadIO m
+githubSearch
```
chshersh: I think this function was previously called `liftGitHubSearchToApp`. In that case, it misses documentation.

chshersh: Also, I believe that this function should be separated into two functions. I think it still makes sense to kepp `liftGitHubSearchToApp` with slight modifications. `lift*` function will perform GitHub query and lift it to our `Monad` and perform some logging. And `githubSearch` will just construct a query from arguments, log query and call `liftGitHubSearchToApp`

src/IW/Sync/Search.hs
```diff
@@ -28,70 +35,117 @@ import qualified Data.Text as T
...
+    -> Day    -- ^ From day
+    -> Day    -- ^ To day
+    -> Int    -- ^ Page
+    -> m [a]
+githubSearch paths queryString from to page = liftIO (executeRequest' $ query paths queries) >>= \case
+    Left err -> throwError $ githubErrToAppErr err
+    Right (SearchResult count vec) -> do
+        log Info $ "Query executed with the following paths and query parameters: " <> show paths <> " " <> show queries
+        log Info $ "Fetched total of " <> show count <> " " <> typeName @a <> "s..."
+        searchLimit <- getSearchRateLimit
```
chshersh: Does querying `searchLimit` decreases the search limit? :slightly_smiling_face: 
Also, it would be nice to perform calls to this function only in `Debug` mode, but this is not currently supported by `co-log` and I see that this is a useful feature...

src/IW/Sync/Search.hs
```diff
@@ -28,70 +35,117 @@ import qualified Data.Text as T
...
+
+    julianDayToIso :: Day -> Text
+    julianDayToIso = fromString . formatTime defaultTimeLocale (iso8601DateFormat Nothing)
+
+getSearchRateLimit
+    :: forall m.
+       ( MonadIO m
+       , WithError m
+       )
+    => m Limits
```
chshersh: Signature for this function can be written in one line. No need to put every argument on a separate line if function fits into single line. Having short functions with short type signatures helps readability better rather than an unnecessary split of type signature into multiple lines.

src/IW/Sync/Search.hs
```diff
@@ -28,70 +35,117 @@ import qualified Data.Text as T
...
+getSearchRateLimit = liftIO rateLimit >>= \case
+    Left err -> throwError $ githubErrToAppErr err
+    Right RateLimit{..} -> pure rateLimitSearch
```
chshersh: This looks like something that can be done by `liftGitHubToApp` function :thinking:

src/IW/Sync/Search.hs
```diff
@@ -28,70 +35,117 @@ import qualified Data.Text as T
...
+    :: forall env m.
+       ( MonadIO m
+       , WithError m
+       , WithLog env m
+       )
+    => Day
+    -> Day
+    -> Int
+    -> m [GitHub.Repo]
+fetchHaskellReposByDate from to page = githubSearch paths queryString from to page
```
chshersh: Can you elaborate on the meaning of `paths` argument? I don't quite understand why it's need and why it's a list of `Text`... Also, it seems like this function can be written shorter and there's no need to introduce variables in `where` clause. Also can eta-reduce the function. Something like:

```haskell
fetchHaskellReposByDate = githubSearch ["search", "repositories"] "language:haskell"
```

src/IW/Sync/Search.hs
```diff
@@ -28,70 +35,117 @@ import qualified Data.Text as T
...
+    Left err -> throwError $ githubErrToAppErr err
+    Right (SearchResult count vec) -> do
+        log Info $ "Query executed with the following paths and query parameters: " <> show paths <> " " <> show queries
+        log Info $ "Fetched total of " <> show count <> " " <> typeName @a <> "s..."
+        searchLimit <- getSearchRateLimit
+        log Info $ "Search rate limit information: " <> show searchLimit
+        pure $ V.toList vec
+  where
+    queries :: [(ByteString, Maybe ByteString)]
+    queries = pagination <> [("q", Just $ encodeUtf8 $ strip queryString <> " " <> dateRange from to)]
```
chshersh: I feel like it will help readability if you introduce helper functions to construct queries. Again, this function seems to do a lot. So it makes to split into separate functions. For example, you can create `buildGitHubQuery` function that takes query string, page number and returns result of a `query` function. 

P.S. I would use `GitHub.query` function qualifed because `query` is too generic name and we have `query` in DB functions, so fully qualifying will help.

src/IW/Sync/Update.hs
```diff
@@ -1,40 +1,86 @@
...
+       , MonadUnliftIO m
+       , WithDb env m
+       , WithLog env m
+       , WithError m
+       )
+    => Repo
+    -> m ()
+syncCategories Repo{..} = do
+    categories <- getCabalCategories repoOwner repoName
+    updateRepoCategories repoOwner repoName categories
```
chshersh: I like when functions are that simple! It's easy to review them :joy:

src/IW/Sync/Update.hs
```diff
@@ -1,40 +1,86 @@
...
+syncRepos oldest recent interval page =
+    if recent == oldest then
+        log I $ "Oldest day reached"
+    else
+        do
+            resCount <- syncReposByDate intervalStart recent page
+            if resCount < 100 then
+                syncRepos oldest nextRecent interval 1
+            else
+                syncRepos oldest recent interval (page + 1)
```
chshersh: I feel like there is a better way to implement this function. It looks like some state machine. Maybe there is a way to analyse arguments and decide what to do next depending on that. This ad-hoc if-then-else logic seems okay for now and it works. So maybe this is okay for now. But I wrote similar code before and it was buggy due to off-by-one errors and various similar arithmetics. So I just want to be more cautious :slightly_smiling_face:

src/IW/Sync/Update.hs
```diff
@@ -1,40 +1,86 @@
...
+    -> Day     -- ^ Most recent day
+    -> Integer -- ^ Interval of days before most recent day
+    -> Int     -- ^ Page
+    -> m ()
+syncRepos oldest recent interval page =
+    if recent == oldest then
+        log I $ "Oldest day reached"
+    else
+        do
+            resCount <- syncReposByDate intervalStart recent page
```
chshersh: Indentation here is too agressive. Can be:

```haskell
else do
    ...
```

src/IW/Sync/Update.hs
```diff
@@ -1,40 +1,86 @@
...
+            if resCount < 100 then
+                syncRepos oldest nextRecent interval 1
+            else
```
chshersh: According to our style guide this should be:

```haskell
if ...
then syncRepos
else syncRepos
```

rashadg1030: The style guide also mentions this style for `do` blocks:
```
    if null answer
        then pure (head choices)
        else pure answer
```

chshersh: @rashadg1030 Yeah, that's true. I think both are fine. Yours is preferrable to be consistent with the style guide

rashadg1030:
@chshersh @vrom911 So I tested the main `syncRepos` (it keeps changing names) function and it seems to be working. A few notes:

1. The code is really messy and needs clean up. Thank you, Dmitrii for pointing out these issues. There are definitely better ways to implement a lot of this stuff. I will fix this tomorrow.

2. I tested the function and was able to fetch all `repos` and insert them into the DB from 01-03-2019 to 29-07-2019 without any hiccups so that was cool. I should've set a timer, but forgot. I think it took around 10 minutes to do the whole job. 

3. I use `threadDelay` to delay the call of the next fetch for one minute when the rate limit remaining value reaches 0.

3. It turns out an interval of 30 days is too big and it crashes the program (this is bad, it shouldn't crash) when more than 1000 search results are returned. I changed the interval to 14 days and it was fine, but this might not be optimal. I need to implement a handler for when more than 1000 search results are returned. It should retry the same query but with a smaller interval.

4. The logging is super helpful, so I think I'm going to go ahead with creating a logging file to see more logs.

5. I also need to find the first Haskell repository on GitHub and the date. 

6. I also still need to add errors and handlers to the functions. I just have logging for now. 

7. Thinking about writing tests as well.

src/IW/Sync/Search.hs
rashadg1030: Good question. Queries for the search limit do not affect the search limit, so it's fine to make calls to it as many times as you want.

src/IW/Sync/Search.hs
rashadg1030: @chshersh Yes for sure, I went too far with this one :sweat_smile:

src/IW/Sync/Search.hs
rashadg1030: The paths arguement is for this function: 
```
query paths queries
```
It constructs a github query given a list of paths and list of query parameter/value pairs. I agree that this function can be a lot simpler. I will fix this.

src/IW/Sync/Update.hs
rashadg1030: Hmmm yes, I agree. I try to avoid if-else logic when possible. I'll come up with something better.

src/IW/Sync/Search.hs
rashadg1030: I forgot to use `Paths` and `QueryString` type synonyms from `github` library. If you look at the new code it should make sense. Maybe still, using `Paths` might be overkill but I prefer it over just a `Text`.

src/IW/Sync/Search.hs
rashadg1030: Hmmm. I'm not sure how to go about this. I see what you're saying I think. I was thinking to remove `liftIO` from this function and make it `IO Limits` but then I think I lose the ability to throw an error right? I could generalize the `liftGithubToApp` function to be able to take the `getSearchRateLimit` function, as I think you suggested, but then I have to change `liftGithubSearchToApp` function to accommodate it and I won't be able to log some useful things for the `fetch...` functions unless I build more functions. Is it worth it? Is there a better way?

src/IW/Sync/Update.hs
rashadg1030: I'm still trying to think about how I can analyze the arguments and then make a decision based on that rather than using if-then-else logic.

rashadg1030:
The `sync...` functions in `Update.hs` can be generalized more. Right now it is just for repos.
Also, hlint (I think that's what it is) recommends me to use `prec` from `Relude.extra.Enum` instead of `pred` on a `Day` value, but `Day` doesn't have an instance of the `Bounded` typeclass so I can't.

src/IW/Sync/Search.hs
```diff
@@ -28,70 +32,109 @@ import qualified Data.Text as T
...
        ( MonadIO m
+       , MonadUnliftIO m
        , WithError m
        , WithLog env m
        , Typeable a
        )
     => IO (Either GitHub.Error (SearchResult a))
     -> m [a]
-liftGitHubSearchToApp githubSearch = liftIO githubSearch >>= \case
+liftGithubSearchToApp githubSearch' = do
```
vrom911: Can we add comments to this logic? It's quite complicated and future us would be thankful for that :slightly_smiling_face:

src/IW/Sync/Search.hs
```diff
@@ -28,70 +32,109 @@ import qualified Data.Text as T
...
+    performSearch = liftIO githubSearch' >>= \case
+        Left err -> throwError $ githubErrToAppErr err
+        Right (SearchResult count vec) -> do
+            log Info $ "Fetched total of " <> show count <> " " <> typeName @a <> "s..."
+            pure $ V.toList vec
+
+    delaySearch :: m [a]
+    delaySearch = do
+        log Info "No more requests remaining. Waiting for one minute..."
+        threadDelay 60000000
```
chshersh: Logic with waiting for 1 minute seems awkward. I feel like there's a better solution..

src/IW/Sync/Search.hs
```diff
@@ -28,70 +32,109 @@ import qualified Data.Text as T
...
-    Right (SearchResult count vec) -> do
-        log Info $ "Fetching total of " <> show count <> " " <> typeName @a <> "s..."
-        pure $ V.toList vec
+    Right RateLimit{..} -> pure rateLimitSearch
+
+-- | Search all repositories built with the Haskell language by page within a date range.
+searchHaskellReposByDate
+    :: Day
+    -> Day
+    -> Int
```
chshersh: Adding comments to the arguments will help :slightly_smiling_face:

src/IW/Sync/Update.hs
```diff
@@ -4,37 +4,82 @@ data and insert it into the database.
...
+        resCount <- syncReposByDate intervalStart recent page
+        if resCount < 100
+            then syncRepos oldest nextRecent interval 1
+            else syncRepos oldest recent interval (page + 1)
+
+    intervalStart :: Day
+    intervalStart = negate interval `addDays` recent
+
+    nextRecent :: Day
+    nextRecent = pred intervalStart
```
chshersh: Yeah, the HLint rule in `relude` should be fixed in that case. For now you can just ignore the rule inside this module.

src/Prelude.hs
```diff
@@ -43,3 +47,6 @@ import qualified Colog (Message, WithLog)
 
 -- | 'Colog.WithLog' alias specialized to 'Message' data type.
 type WithLog env m = Colog.WithLog env Colog.Message m
+
+julianDayToIso :: Day -> Text
+julianDayToIso = fromString . formatTime defaultTimeLocale (iso8601DateFormat Nothing)
```
chshersh: Again, this function is not going to be used in every module. So it's better to put it into a separate module like `IW.Time`

src/IW/Sync/Search.hs
rashadg1030: @chshersh I was thinking I could just recursively keep checking the search rate limit (it doesn’t lower the search count) and then make a request when the search limit resets back to 10.

src/IW/Sync/Search.hs
```diff
@@ -1,97 +1,139 @@
...
        , Typeable a
        )
     => IO (Either GitHub.Error (SearchResult a))
     -> m [a]
-liftGitHubSearchToApp githubSearch = liftIO githubSearch >>= \case
+liftGithubSearchToApp githubSearch' = do
+    searchLimit <- getSearchRateLimit
+    if limitsRemaining searchLimit > 0
+        then performSearch
+        else liftGithubSearchToApp githubSearch'
```
chshersh: Let's for now go with the simpler solution: just do `throwError` is we reached the limit. We will configure restart strategy separately

src/IW/Sync/Search.hs
```diff
@@ -1,97 +1,139 @@
...
        ( MonadIO m
+       , MonadUnliftIO m
        , WithError m
        , WithLog env m
        , Typeable a
        )
     => IO (Either GitHub.Error (SearchResult a))
     -> m [a]
-liftGitHubSearchToApp githubSearch = liftIO githubSearch >>= \case
+liftGithubSearchToApp githubSearch' = do
```
chshersh: Let's rename `githubSearch'` to `searchAction`

src/IW/Sync/Search.hs
```diff
@@ -1,97 +1,139 @@
...
+githubSearch paths queryString from to page = executeRequest' $ buildGithubQuery paths queryString from to page
+
+-- | Function for building a GitHub search query value.
+buildGithubQuery
+    :: Paths -- ^ URL paths
+    -> Text  -- ^ Query string for GitHub search
+    -> Day   -- ^ First day of date range that values were created in
+    -> Day   -- ^ Last day of date range that values were created in
+    -> Int   -- ^ Number of page to be returned
+    -> GitHub.GenRequest 'GitHub.MtJSON 'GitHub.RO a
```
chshersh: Heh, these are very ugly types from the `github` library. But nothing we can do here

src/IW/Time.hs
```diff
@@ -0,0 +1,18 @@
...
+       , getToday
+       , julianDayToIso
+       ) where
+
+import Data.Time (Day (..), getCurrentTime, utctDay)
+import Data.Time.Format (formatTime, defaultTimeLocale, iso8601DateFormat)
+
+
+firstHaskellRepoCreated :: Day
+firstHaskellRepoCreated = ModifiedJulianDay 58211
```
chshersh: Can you, please, add a comment to this constant what is this date in human calendar?

src/IW/Time.hs
```diff
@@ -0,0 +1,18 @@
...
+
+getToday :: IO Day
+getToday = utctDay <$> getCurrentTime
+
+julianDayToIso :: Day -> Text
```
chshersh: Also would be nice to have an example of the text result this function returns in the documentation

src/IW/Sync/Update.hs
```diff
@@ -28,13 +31,68 @@ fetchAndUpsertRepos
...
   where
-    fetchAndUpdateCategories :: Repo -> m ()
-    fetchAndUpdateCategories Repo{..} = do
-        categories <- getCabalCategories repoOwner repoName
-        updateRepoCategories repoOwner repoName categories
+    intervalStart :: Day
+    intervalStart = negate interval `addDays` recent
+
+    nextRecent :: Day
+    nextRecent = pred intervalStart
```
chshersh: Regarding HLint warning on the CI: you can disable this warning in this particular file for now. You can see how it's done here:

* https://github.com/kowainik/kowainik.github.io/blob/0f889b8c85cccc5296bd1f4a5c7499f219d77c20/site.hs#L28

And, if you have time, you can also open the corresponding issue to the `relude` and we will fix it in the next release :slightly_smiling_face:

src/IW/Sync/Update.hs
```diff
@@ -28,13 +31,68 @@ fetchAndUpsertRepos
        , WithError m
        )
     => m ()
-fetchAndUpsertRepos = do
-    gitHubRepos <- fetchAllHaskellRepos
+syncRepos = do
+    today <- liftIO getToday
+    sync syncReposByDate firstHaskellRepoCreated today 15 1
```
chshersh: I feel like this function can be improved a lot. But at the same time I feel that if we spend a lot of time improving this function, we will never get to the rest of the project :laughing: From the top of my head:

1. We can use `newtype`s here
2. For now you do GitHub query for every batch of issues, but you can fetch all issues first and then insert all at once. Though, not sure that it will be faster.

But I think that the current implementation is good enough.

src/IW/Time.hs
vrom911: And also, I'm so curious, what is the first Haskell library at GitHub?

rashadg1030: The first Haskell repo was created on 2008-04-03!

src/IW/Sync/Update.hs
rashadg1030: Yes! I agree. I think I've finally realized a better way to do this.

rashadg1030:
@chshersh I refactored the Update and Search modules one last time. I thought this would be better than the other way. A lot of the stuff that Update was doing, I pushed to the Search module. It makes more sense. Let me know what you think. I still need to improve some of the functions and clean up some naming and other minor stuff. To be honest I have trouble deciding on how many functions I want to divide a task into lol and if it should be in the IO monad or our App monad. I get the urge to lift everything into the App monad immediately, but we don’t want that right? Anyways, this is the last time I will touch this branch after I clean it up a bit tonight. I’m gonna move on to other stuff

src/IW/Time.hs
chshersh: @rashadg1030 Do you have a link to this repo? :slightly_smiling_face:

src/IW/App/Error.hs
```diff
@@ -207,6 +208,9 @@ dbError = InternalError . DbError
 dbNamedError :: PgNamedError -> AppErrorType
 dbNamedError = InternalError . DbNamedError
 
+githubHTTPError :: Text -> AppErrorType
+githubHTTPError = GitHubError . HTTPError
```
chshersh: It should be `githubHttpError` according to our style guide

src/IW/Sync/Update.hs
```diff
@@ -14,27 +14,36 @@ import IW.App (WithError)
...
        , WithLog env m
        , WithError m
        )
-    => m ()
-fetchAndUpsertRepos = do
-    gitHubRepos <- fetchAllHaskellRepos
-    let repos = map fromGitHubRepo gitHubRepos
+    => Integer -- ^ Starting interval
+    -> m ()
+syncRepos interval = do
```
chshersh: Is this `interval` a step for syncing?

src/IW/Time.hs
```diff
@@ -0,0 +1,20 @@
...
+       ( firstHaskellRepoCreated
+       , getToday
+       , julianDayToIso
+       ) where
+
+import Data.Time (Day (..), getCurrentTime, utctDay)
+import Data.Time.Format (formatTime, defaultTimeLocale, iso8601DateFormat)
+
+
+-- | This day is equilvalent to 2008-04-03.
```
chshersh: Hmm. this is a very old repo... But I guess it should be fine. Could you try to sync up until this day and tell how much time it requires?

src/IW/Sync/Update.hs
```diff
@@ -14,27 +14,36 @@ import IW.App (WithError)
...
+    repos <- searchAllHaskellRepos interval
     upsertRepos repos
-    mapConcurrently_ fetchAndUpdateCategories repos
-  where
-    fetchAndUpdateCategories :: Repo -> m ()
-    fetchAndUpdateCategories Repo{..} = do
-        categories <- getCabalCategories repoOwner repoName
-        updateRepoCategories repoOwner repoName categories
+    mapConcurrently_ syncCategories repos
```
chshersh: `syncRepos` now look simple. But I guess the main complexity is hidden inside `searchAllHaskellRepos`.

src/IW/Sync/Search.hs
```diff
@@ -1,97 +1,195 @@
...
+       | count < 1000 -> do
+            remResults <- foldMapM (githubSearch' paths properties startDay recent) [2..10]
+            githubSearchAll paths properties (pred startDay) interval (acc <> V.toList vec <> remResults)
+        -- | Otherwise, call the function with the same arguments but a smaller interval.
+       | otherwise    -> githubSearchAll paths properties recent (pred interval) acc
 
--- | Search all repositories built with the Haskell language.
-searchHaskellRepos :: IO (Either GitHub.Error (SearchResult GitHub.Repo))
-searchHaskellRepos = searchRepos "language:haskell"
+githubSearch'
```
chshersh: I personally don't like `'` in names just to have different functions. Can we rename this function or `githubSearch` to something else to have clear difference between functions?

src/IW/Time.hs
rashadg1030: @chshersh I will try that rn

src/IW/Sync/Update.hs
rashadg1030: @chshersh yeah I thought this was better. I felt like all that functionality I had in syncRepos belonged to the actual search function instead.

src/IW/Sync/Update.hs
rashadg1030: @chshersh This interval is used for the search functions. The interval is for determining the date interval. Look at the search functions for a better understanding, but basically it stands for many days worth of issues/repos we’re fetching. The comments on some of the functions need to be updated. I’m doing that rn.

src/IW/Sync/Search.hs
rashadg1030: I agree completely. Couldn’t think of a name so that’s just a placeholder until I come up with something better. I hate getting stuck st naming functions😂

rashadg1030:
@chshersh I tried testing the function forgot about how do we handle reaching the API rate limit. Before, I was using `threadDelay` for a minute to handle this problem, but you recommended against it. What alternatives do I have? I considered just recursively calling until the search goes through, but that might be inefficient. For now, an error is thrown if the limit is reached.

```
-- | This function is for lifting a GitHub search action to the App monad.
liftGithubSearchToApp
    :: forall a m.
       ( MonadIO m
       , MonadUnliftIO m
       , WithError m
       )
    => IO (Either GitHub.Error (SearchResult a))
    -> m (SearchResult a)
liftGithubSearchToApp searchAction = do
    searchLimit <- getSearchRateLimit
    if limitsRemaining searchLimit > 0
        then liftIO searchAction >>= \case
            Left err -> throwError $ githubErrToAppErr err
            Right searchRes -> pure searchRes
        else throwError $ githubHTTPError "GitHub Search API limit reached."
```

rashadg1030:
@chshersh Okay so my wi-fi isn't working right now, so I can't get an accurate test. I tried it for a bit and it was super slow (using my iPhone as a hotspot). I will try tomorrow or maybe someone can test it on their machine. One thing I noticed is that I need more logging so I will add that right now.

rashadg1030:
@chshersh @vrom911 Some quick napkin math after testing `syncRepos`. It took 20 minutes to sync 8 months worth of repos. That means it would take 1 hour to sync 2 years worth of repos. To get to the oldest Haskell repo, which is roughly 11 years and 4 months ago, it would take about 6 hours...Should we put a cap on how far we go back?

rashadg1030:
@chshersh @vrom911 I'm adding no more to this PR, but let me know if there is a minor improvement that can be made. I am satisfied for now with the `Search` and `Update` modules. Let me know what you think. I still feel like it could be better, but I got to move on. I got it in a good place where I can implement the `Search` and `Update` functions for issues pretty easily I think. Let's find out!

And I wanted to get the link to the binary search method for optimizing the GitHub search. I think @chshersh posted it in a past discussion, but I can't find it...I'm not going to implement it any time soon, but I would like to open an issue on it.

src/IW/Sync/Search.hs
```diff
@@ -1,97 +1,214 @@
...
+    log I $ "Current search limit information: " <> show searchLimit
+    if limitsRemaining searchLimit > 0
+        then do
+            log D $ "Searching GitHub API with following request: " <> showGithubQuery paths properties from to page
+            liftIO (githubSearch paths properties from to page) >>= \case
+                Left err -> throwError $ githubErrToAppErr err
+                Right searchRes -> pure searchRes
+        else do
+            log I "API limit reached. Delaying search..."
+            threadDelay 60000000
```
chshersh: Let's just throw an error here. The limits for GitHub API is 5000 per hour. So waiting one minute doesn't really help... If we hit the limit, we will handle it separately. You can create the issue to handle cases when we reach rate limit.

PR #129, opened 2019-08-08
[#99] Add logging and error handling to `parseUserData` function

### LINKED ISSUE #99: Add logging and error handling to `fromGitHubIssue` and `parseIssueUserData` functions
For now, the `fromGitHubIssue` and related `parseIssueUserData` function returns `Maybe` values. We get `Nothing` when an error occurs, but it's a good idea to add error handling and logging for parse errors in the future.

rashadg1030:
Resolves #99

### DIFF
--- a/src/IW/Sync/Search.hs
+++ b/src/IW/Sync/Search.hs
@@ -205,7 +205,7 @@ fromGitHubRepo githubRepo = Repo
     }
 
 -- | Convert a value of the @GitHub.Issue@ type to a value of our own @Issue@ type.
-fromGitHubIssue :: WithError m => GitHub.Issue -> m Issue
+fromGitHubIssue :: (WithError m, WithLog env m) => GitHub.Issue -> m Issue
 fromGitHubIssue githubIssue = do
     (issueRepoOwner, issueRepoName) <- parseUserData $ GitHub.issueUrl githubIssue
     pure Issue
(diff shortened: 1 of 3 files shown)


chshersh:
@rashadg1030 Would it be possible to keep only relevant changes in this PR without depending on the PR about sync with GitHub? It looks like you've branched out from other existing PR. But according to our workflow, you should always branch out from `master` unless your PR depends on the current PR. Because of that, it's harder to review this PR independently. And, as you noticed, #127 is huge and complicated already to also be reviewed as a part of another PR.

src/IW/Sync/Search.hs
```diff
@@ -229,10 +222,9 @@ fromGitHubIssue githubIssue = do
 Parsing the URL @https://api.github.com/repos/owner/name/issues/1@ should return
 @Just (RepoOwner "owner", RepoName "name")@.
 -}
-parseUserData :: GitHub.URL -> Maybe (RepoOwner, RepoName)
-parseUserData (URL url) =
-    T.stripPrefix "https://api.github.com/repos/" url
-    >>= splitOwnerAndName
+parseUserData :: WithError m => GitHub.URL -> m (RepoOwner, RepoName)
+parseUserData (URL url) = notFoundOnNothing
```
vrom911: I guess we can provide a more verbose log in case of `Nothing` as well. As `splitOwnerAndName` is a pure function, we would know the reason for failed parsing for sure.

rashadg1030: @vrom911 Excuse me, but I'm not sure if I understood what you meant. I added a more verbose log message, but I'm not sure if it is helpful. Did you mean a log message for `stripPrefix`?

PR #130, opened 2019-08-08
[#110] Remove Internal/External error wrappers used for AppErrorType

### LINKED ISSUE #110: Remove Internal/External error wrappers used in `AppErrorType`

rashadg1030:
Resolves #110

### DIFF
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -94,15 +94,6 @@ data AppError = AppError
 
 -- | App errors type.
 data AppErrorType
-    = InternalError IError
-    | GitHubError GError
-    | UrlDownloadFailed Url
-    deriving (Show, Eq)
-
-{- | The internal errors that can be thrown. These errors are meant to be
-handled within the application and cover exceptional circumstances/coding errors.
--}
-data IError
     {- | General not found. -}
     = NotFound
     {- | Some exceptional circumstance has happened stop execution and return.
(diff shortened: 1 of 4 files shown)


src/IW/App/Error.hs
```diff
@@ -126,90 +117,90 @@ data IError
...
-dbError = InternalError . DbError
+dbError = DbError
 
 dbNamedError :: PgNamedError -> AppErrorType
-dbNamedError = InternalError . DbNamedError
+dbNamedError = DbNamedError
 
 githubHttpError :: Text -> AppErrorType
-githubHttpError = GitHubError . HttpError
+githubHttpError = GithubHttpError
```
vrom911: I'm not sure that these helper functions are so helpful now when we inlined the error data type. Do you think we should keep them, or just use the constructors?

rashadg1030: @vrom911 We should just use constructors imo

PR #132, opened 2019-08-08
[#45] Add async worker for synchronizing database

### LINKED ISSUE #45: Add async worker for populating DB

rashadg1030:
Resolves #45

### DIFF
--- a/src/IW/Sync/Update.hs
+++ b/src/IW/Sync/Update.hs
@@ -68,14 +68,6 @@ syncCategories
        )
     => Repo
     -> m ()
-<<<<<<< HEAD
 syncCategories repo = do
     categories <- getCabalCategories repo
     updateRepoCategories repo categories
-=======
-syncCategories Repo{..} = do
-    log I "Updating repository's categories..."
-    categories <- getCabalCategories repoOwner repoName
-    updateRepoCategories repoOwner repoName categories
-    log I "Categories successfully updated"
->>>>>>> c36be34... [#45] Add sync function for synchronizing all repositories and issues
--- a/src/IW/Sync/Update.hs
+++ b/src/IW/Sync/Update.hs
@@ -4,23 +4,35 @@ data and insert it into the database.
 -}
 
 module IW.Sync.Update
-       ( syncRepos
+       ( syncCache
        ) where
 
 import Control.Monad.IO.Unlift (MonadUnliftIO)
-import UnliftIO.Async (mapConcurrently_)
+import UnliftIO.Async (async, mapConcurrently_, wait)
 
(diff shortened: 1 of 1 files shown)


src/IW/Sync/Update.hs
```diff
@@ -30,13 +42,22 @@ syncRepos
        )
     => Integer -- ^ The starting date interval used in the search function
     -> m ()
-syncRepos interval = do
+syncWithGithub interval = do
+    log I "Starting synchronization of all repositories and issues..."
     today <- liftIO getToday
+    log I "Searching for all Haskell repositories..."
     repos <- searchAllHaskellRepos today interval
+    log I "Upserting repositories into the database..."
```
chshersh: Logging is very useful in this function! And we definitely should have it. It's just unfortunate that because of calls to logging functions the code becomes less readable and all steps are not apparent immediately... It's an open question, how to improve the situation though. But at least, these logging messages serve as a good documentation

rashadg1030: @chshersh Yes, I agree. I thought the same, but I think it’s definitely worth it. This was my first time using any logging library but I really enjoy how easy and straightforward it is to use. Very user friendly.

src/IW/Sync/Update.hs
```diff
@@ -4,23 +4,38 @@ data and insert it into the database.
...
+       , WithDb env m
+       , WithLog env m
+       , WithError m
+       )
+    => Integer
+    -> m ()
+syncCache interval = forever $ syncWithGithub interval `catchError` syncErrHandler
+  where
+    syncErrHandler :: AppErrorType -> m a
+    syncErrHandler = undefined
```
rashadg1030: @chshersh @vrom911 Okay so once I implement this handler I think this PR is done. How should we handle the errors? We don't want it to stop syncing if an error is encountered, so do I have to think of some way to resume the sync process from where the error was thrown? Or is it simpler than that?

chshersh: I will think about the problem :thinking:

PR #136, opened 2019-08-18
[#114] Implement proper pagination

### LINKED ISSUE #9: Write Servant endpoint that returns Haskell repositories
After #8 is done

### LINKED ISSUE #114: Implement proper pagination
In our case pagination should be implemented on the backend because we have a lot of issues. Every query that returns issues, should end with the following:

```sql
LIMIT ISSUES_PER_PAGE
OFFSET (PAGE_NUMBER * ISSUES_PER_PAGE)
```

Where:

1. `ISSUES_PER_PAGE` is hardcoded constant (like 100)
2. `PAGE_NUMBER`: this is what frontend gives to backend

rashadg1030:
Resolves #114 
Resolves #9

### DIFF
--- a/src/IW/Server.hs
+++ b/src/IW/Server.hs
@@ -3,15 +3,23 @@ module IW.Server
        , server
        ) where
 
+import Servant.API ((:<|>))
 import Servant.API.Generic (toServant)
 import Servant.Server (Server, hoistServer)
 
 import IW.App (AppEnv)
 import IW.Effects.Log (runAppAsHandler)
-import IW.Server.Issue (IssueAPI, issueServer)
+import IW.Server.Issue (IssuesAPI, issuesHandler)
+import IW.Server.Repo (ReposAPI, reposHandler)
 
 
-type API = IssueAPI
+type API = ReposAPI :<|> IssuesAPI
 
 server :: AppEnv -> Server API
-server env = hoistServer (Proxy @API) (runAppAsHandler env) (toServant issueServer)
+server env = hoistServer (Proxy @API) (runAppAsHandler env) (toServant iwServer)
+
+iwServer :: IssuesSite AppServer
+iwServer = IssuesSite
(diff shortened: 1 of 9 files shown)


src/IW/Server.hs
```diff
@@ -3,15 +3,23 @@ module IW.Server
...
 
 import IW.App (AppEnv)
 import IW.Effects.Log (runAppAsHandler)
-import IW.Server.Issue (IssueAPI, issueServer)
+import IW.Server.Issue (IssuesAPI, issuesHandler)
+import IW.Server.Repo (ReposAPI, reposHandler)
 
 
-type API = IssueAPI
+type API = ReposAPI :<|> IssuesAPI
```
vrom911: I thinл that sticking to the same scheme and conventions would make sense here as well. Instead, you can do something like this and it would be much more consistent and structured:

```haskell
data IwSite route = IwSite
    { iwIssuesRoute :: route :- IssuesApi
    , iwIReposRoute :: route :- ReposApi
    } deriving (Generic)

type IwApi = ToApi IwSite

iwServer :: IwSite AppServer
iwServer = IwSite
    { iwIssuesRoute = toServant ...
    , iwReposRoute = toServant ...
    }
```

and also, in this case you don't need to change `Server.Issue` and `Server.Repo` servers and use them instead of handlers here.

rashadg1030: @vrom911 Got it.

rashadg1030: @vrom911 Yeah this makes sense, so do you ever use the `(:<|>)` operator when using `servant`?

vrom911: We don't use it anywhere, to be honest. `servant-generic` (which is currently merged to `servant` itself and which we are using everywhere) technique is much nicer to my taste. But it's possible to use for some small stuff.

PR #142, opened 2019-08-22
[#102] Add handling of cabal file parse errors

### LINKED ISSUE #102: Add handling of cabal file parser errors

rashadg1030:
Resolve #102

### DIFF
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -5,6 +5,7 @@ module IW.App.Error
        , AppErrorType (..)
        , AppException (..)
        , CabalPError (..)
+       , CabalErrorInfo (..)
        , WithError
        , githubErrToAppErr
        , throwError
@@ -123,7 +124,10 @@ data AppErrorType
     | UrlDownloadFailed Url
     deriving (Show, Eq)
 
-type CabalErrorInfo = (Maybe Version, [CabalPError])
+data CabalErrorInfo = CabalErrorInfo
+    { cabalVersion :: Maybe Version
+    , cabalPErrors :: [CabalPError]
+    } deriving (Show, Eq)
 
 {- | A wrapper around the 'PError' type from the @Cabal@ library.
 This is needed to implement an instance of 'Eq' for 'PError' without
@@ -136,7 +140,7 @@ instance Eq CabalPError where
     CabalPError (PError pos1 str1) == CabalPError (PError pos2 str2) =
         pos1 == pos2 && str1 == str2
 
--- | Map the @github@ library's 'Error' type into 'AppErrorType'.
+-- | Map the @github@ library's 'GitHub.Error' type into 'AppErrorType'.
 githubErrToAppErr :: GitHub.Error -> AppErrorType
 githubErrToAppErr = \case
     GitHub.HTTPError e    -> GithubHttpError $ show e
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -108,7 +108,7 @@ data AppErrorType
     | DbError Text
     {- | Data base named parameters errors. -}
     | DbNamedError PgNamedError
-    {- | -}
+    {- | An error occurred while attempting to parse a @.cabal@ file. -}
     | CabalParseError (Maybe Version, [CabalPError])
     {- | A HTTP error occurred using the @github@ library.
     The actual caught error is included. -}
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -87,7 +87,7 @@ data AppError = AppError
 data AppErrorType
     {- | General not found. -}
     = NotFound
-    {- | Some exceptional circumstance has happened stop execution and return.
+    {- | Some exceptional circumstance has happened to stop execution and return.
     Optional text to provide some context in server logs.
     -}
     | ServerError Text
@@ -130,7 +130,7 @@ instance Eq CabalPError where
     (CabalPError (PError pos1 str1)) == (CabalPError (PError pos2 str2)) =
         pos1 == pos2 && str1 == str2
 
--- | Map the @github@ library's @Error@ type into AppErrorType.
+-- | Map the @github@ library's @Error@ type into @AppErrorType@.
 githubErrToAppErr :: GitHub.Error -> AppErrorType
 githubErrToAppErr = \case
     GitHub.HTTPError e    -> GithubHttpError $ show e
@@ -138,7 +138,7 @@ githubErrToAppErr = \case
     GitHub.JsonError msg  -> GithubJsonError msg
     GitHub.UserError msg  -> GithubUserError msg
 
--- | Map 'AppError' into a HTTP error code.
+-- | Map @AppError@ into a HTTP error code.
 toHttpError :: AppError -> Servant.ServerError
 toHttpError AppError{..} = case appErrorType of
     NotFound               -> err404
@@ -176,20 +176,20 @@ isInvalid _           = False
 -- Helpers
 ----------------------------------------------------------------------------
 
--- | Extract the value from a maybe, throwing the given 'AppError' if
--- the value does not exist
+-- | Extract the value from a @Maybe a@, throwing the given @AppError@ if
+-- the value does not exist.
 throwOnNothing :: WithError m => AppErrorType -> Maybe a -> m a
 throwOnNothing err = withFrozenCallStack . maybe (throwError err) pure
 
--- | Extract the value from a 'Maybe' in @m@, throwing the given 'AppError' if
--- the value does not exist
+-- | Extract the value from a @Maybe a@ in @m@, throwing the given @AppError@ if
+-- the value does not exist.
 throwOnNothingM :: WithError m => AppErrorType -> m (Maybe a) -> m a
 throwOnNothingM err action = withFrozenCallStack $ action >>= throwOnNothing err
 
--- | Similar to 'throwOnNothing' but throws a 'NotFound' if the value does not exist
+-- | Similar to @throwOnNothing@ but throws a @NotFound@ if the value does not exist.
 notFoundOnNothing :: WithError m => Maybe a -> m a
 notFoundOnNothing = withFrozenCallStack . throwOnNothing NotFound
 
--- | Similar to 'throwOnNothingM' but throws a 'NotFound' if the value does not exist
+-- | Similar to @throwOnNothingM@ but throws a @NotFound@ if the value does not exist.
 notFoundOnNothingM :: WithError m => m (Maybe a) -> m a
 notFoundOnNothingM = withFrozenCallStack . throwOnNothingM NotFound
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -4,6 +4,7 @@ module IW.App.Error
        ( AppError (..)
        , AppErrorType (..)
        , AppException (..)
+       , CabalPError (..)
        , WithError
        , githubErrToAppErr
        , throwError
--- a/src/IW/App/Error.hs
+++ b/src/IW/App/Error.hs
@@ -24,6 +24,8 @@ module IW.App.Error
 
 import Control.Monad.Except (MonadError)
 import Data.CaseInsensitive (foldedCase)
+import Distribution.Parsec.Common (PError (..))
+import Distribution.Types.Version (Version)
 import GHC.Stack (SrcLoc (SrcLoc, srcLocModule, srcLocStartLine))
 import Network.HTTP.Types.Header (HeaderName)
 import PgNamed (PgNamedError)
@@ -98,13 +100,15 @@ data AppErrorType
     -}
     | MissingHeader HeaderName
     {- | An authentication header that was required was provided but not in a
-    format that the server can understand
+    format that the server can understand.
     -}
     | HeaderDecodeError Text
-    -- | Data base specific errors
+    {- | Data base specific errors. -}
     | DbError Text
-    -- | Data base named parameters errors.
+    {- | Data base named parameters errors. -}
     | DbNamedError PgNamedError
+    {- | -}
+    | CabalParseError (Maybe Version, [CabalPError])
     {- | A HTTP error occurred using the @github@ library.
     The actual caught error is included. -}
     | GithubHttpError Text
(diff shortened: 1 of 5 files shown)


src/IW/App/Error.hs
```diff
@@ -118,15 +123,22 @@ data AppErrorType
     | UrlDownloadFailed Url
     deriving (Show, Eq)
 
--- | Map the @github@ library's @Error@ type into AppErrorType.
+newtype CabalPError = CabalPError { unCabalPError :: PError }
```
vrom911: I think that `CabalParseError` is not that long, but it brings much more into the context immediately. Anyway, the documentation would be handy here.

Also, by the style guide the indentation should be modified a bit:
```suggestion
newtype CabalPError = CabalPError 
    { unCabalPError :: PError 
    } 
```

src/IW/App/Error.hs
```diff
@@ -118,15 +123,22 @@ data AppErrorType
...
+    deriving (Show)
```
vrom911:
```suggestion
    deriving newtype (Show)
```

src/IW/App/Error.hs
```diff
@@ -118,15 +123,22 @@ data AppErrorType
...
+
+instance Eq CabalPError where
+    (CabalPError (PError pos1 str1)) == (CabalPError (PError pos2 str2)) =
+        pos1 == pos2 && str1 == str2
```
vrom911: Isn't it the same as what the derived instance would look like?

src/IW/App/Error.hs
```diff
@@ -118,15 +123,22 @@ data AppErrorType
...
+-- | Map the @github@ library's @Error@ type into @AppErrorType@.
 githubErrToAppErr :: GitHub.Error -> AppErrorType
 githubErrToAppErr = \case
     GitHub.HTTPError e    -> GithubHttpError $ show e
     GitHub.ParseError msg -> GithubParseError msg
     GitHub.JsonError msg  -> GithubJsonError msg
     GitHub.UserError msg  -> GithubUserError msg
 
--- | Map 'AppError' into a HTTP error code.
+-- | Map @AppError@ into a HTTP error code.
```
vrom911: I guess `AppError` is the correct reference here. Any reason, you don't want to link to the data type in docs?

src/IW/App/Error.hs
```diff
@@ -163,20 +176,20 @@ isInvalid _           = False
 -- Helpers
 ----------------------------------------------------------------------------
 
--- | Extract the value from a maybe, throwing the given 'AppError' if
--- the value does not exist
+-- | Extract the value from a @Maybe a@, throwing the given @AppError@ if
```
vrom911: The same about `AppError` here. `'AppError'` creates a haddock link to the data type while `@AppError@` is just the code highlight in the documentation.

vrom911:
```suggestion
-- | Extract the value from a 'Maybe', throwing the given @AppError@ if
```

src/IW/App/Error.hs
```diff
@@ -163,20 +176,20 @@ isInvalid _           = False
...
+-- the value does not exist.
 throwOnNothing :: WithError m => AppErrorType -> Maybe a -> m a
 throwOnNothing err = withFrozenCallStack . maybe (throwError err) pure
 
--- | Extract the value from a 'Maybe' in @m@, throwing the given 'AppError' if
--- the value does not exist
+-- | Extract the value from a @Maybe a@ in @m@, throwing the given @AppError@ if
```
vrom911: The same goes here and further in this module

src/IW/Effects/Cabal.hs
```diff
@@ -15,15 +15,16 @@ module IW.Effects.Cabal
...
+import IW.App (App (..), AppErrorType (..), CabalPError (..), WithError,
+               throwError)
 import IW.Core.Repo (Repo (..), Category (..))
-import IW.Core.Url (Url (..))
-import IW.Effects.Download (MonadDownload (..), downloadFileMaybe)
+import IW.Effects.Download (MonadDownload (..))
 
 
--- | Describes a monad that returns @[Category]@ given a @RepoOwner@ and @RepoName@.
+-- | Describes a monad that returns @[Category]@ given a @Repo@.
```
vrom911: All these types can be links (using `''`) in docs.

src/IW/Effects/Cabal.hs
```diff
@@ -32,26 +33,31 @@ instance MonadCabal App where
...
 getCabalCategoriesImpl Repo{..} = do
-    maybeCabalFile <- downloadFileMaybe repoCabalUrl
-    case maybeCabalFile >>= parseGenericPackageDescriptionMaybe of
-        Nothing -> do
-            log W $ "Couldn't parse file downloaded from " <> unUrl repoCabalUrl
-            pure []
-        Just genPkgDescr -> do
-            log I $ "Successfully parsed file downloaded from " <> unUrl repoCabalUrl
+    cabalFile <- downloadFile repoCabalUrl
+    let result = runParseResult $ parseGenericPackageDescription cabalFile
```
vrom911: That would be helpful if you pattern match on the tuple here, like:

```haskell
let (warnings, somethingElse) = ...
```
if you add the comment about what this result is it will do as well :slightly_smiling_face:

src/IW/App/Error.hs
rashadg1030: @vrom911 I was ignorant of how haddock works and replaced the `'...'` with `@...@` for no good reason. I'm going to change it back. I thought the `'...'` were for something else.

vrom911: Oh, no worries! Here you can check what `''` means in haddock:
* https://haskell-haddock.readthedocs.io/en/latest/markup.html#hyperlinked-identifiers

src/IW/App/Error.hs
rashadg1030: @vrom911 I thought so too, but for some reason the `PError` type from the `Cabal` package isn't an instance of `Eq`. So I thought that it would still be fine I could just implement it. But when I implement an `Eq` instance for `PError` it compiles but gives me the `orphan-instances` warning. I learned that this happens when the instance isn't declared in the original module that the type comes from. So I had to use a newtype wrapper, in this case `CabalPError`, and implement the instance manually. Let me know if there is a better way to do this.

src/IW/App/Error.hs
rashadg1030: @vrom911 I'm gonna change this back. I'm just finding out what `'...'` does. My mistake I need to learn the haddock syntax because this is a super useful tool. Anyways, should I also link to functions as well? I use `@...@` a lot, but I think I'm going to replace a lot of them. Before the final submit I'm parsing through each module and making sure the documentation is up to date, so let me know if this is a good idea. In what cases should `@...@` be used? It seems to me that you would always want to link to the function/data type docs.

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 Oooh yes, I can do that. Much nicer.

src/IW/App/Error.hs
rashadg1030: @vrom911 I'm sorry I didn't quite understand the first sentence. Should I choose a different name for `CabalParseError`? I wrote some docs for it.

src/IW/App/Error.hs
```diff
@@ -118,7 +123,18 @@ data AppErrorType
...
--- | Map the @github@ library's @Error@ type into AppErrorType.
+{- | A wrapper around the 'PError' type from the @Cabal@ library.
+This is needed to implement an instance of 'Eq' for 'PError' without
+triggering the @orphan-instances@ warning. -}
+newtype CabalPError = CabalPError 
+    { unCabalPError :: PError 
+    } deriving newtype (Show)
+
+instance Eq CabalPError where
+    (CabalPError (PError pos1 str1)) == (CabalPError (PError pos2 str2)) =
```
chshersh: Outer parenthesis are redundant when defining operator. So we can make code less noisy
```suggestion
    CabalPError (PError pos1 str1) == CabalPError (PError pos2 str2) =
```

src/IW/App/Error.hs
```diff
@@ -98,13 +101,15 @@ data AppErrorType
...
     -}
     | HeaderDecodeError Text
-    -- | Data base specific errors
+    {- | Data base specific errors. -}
     | DbError Text
-    -- | Data base named parameters errors.
+    {- | Data base named parameters errors. -}
     | DbNamedError PgNamedError
+    {- | An error occurred while attempting to parse a @.cabal@ file. -}
+    | CabalParseError (Maybe Version, [CabalPError])
```
chshersh: I would prefer to put ` (Maybe Version, [CabalPError])` in a separate data type, like `ParseErrorInfo` for readability. This particular tuple doesn't look readable enough.

src/IW/App/Error.hs
chshersh: @rashadg1030 I think @vrom911 means that `CabalPError` should be renamed to `CabalParseError` and I agree with that. It's more readable. You will have collisions with the `AppErrorType` constructor but you can rename the constructor to just `CabalError` to avoid clashes.

src/IW/App/Error.hs
vrom911: Got it, thanks for the clarification!

src/IW/App/Error.hs
vrom911: Yes, your assumption is correct, for all the functions, data types etc. you would want to make a link to it, that is super useful! `@...@` is used for code blocks, for example, you may want to write @2 + 2@, or else.

src/IW/App/Error.hs
rashadg1030: @chshersh Ok nice. Could this be discovered by hlint?

src/IW/App/Error.hs
rashadg1030: @chshersh Okay got it.

rashadg1030: @chshersh I chose to name it `CabalParseErrorInfo` but it's kinda long. I just thought that `ParseErrorInfo` was kind of vague.

rashadg1030: @chshersh Changed to `CabalErrorInfo`

rashadg1030:
@chshersh @vrom911 Sorry for the other names in the commit history. I'm using my girlfriend's machine because my laptop screen broke...

src/IW/App/Error.hs
```diff
@@ -118,7 +123,20 @@ data AppErrorType
...
+triggering the @orphan-instances@ warning. -}
+newtype CabalPError = CabalPError
+    { unCabalPError :: PError
+    } deriving newtype (Show)
+
+instance Eq CabalPError where
+    CabalPError (PError pos1 str1) == CabalPError (PError pos2 str2) =
+        pos1 == pos2 && str1 == str2
+
+-- | Map the @github@ library's 'Error' type into 'AppErrorType'.
```
vrom911:
```suggestion
-- | Map the @github@ library's 'GitHub.Error' type into 'AppErrorType'.
```
(it would be displayed as just `Error`, otherwise, it is not in scope.

src/IW/Effects/Cabal.hs
```diff
@@ -32,34 +33,40 @@ instance MonadCabal App where
 
 type WithCabal env m = (MonadDownload m, WithLog env m, WithError m)
 
-{- | This function may throw anyone of the errors inherited by the use of @downloadFile@
-defined in @IW.Effects.Download@. We are using @parseGenericPackageDescriptionMaybe@
-which will return @Nothing@ on an unsuccessful parse.
+{- | This function takes a 'Repo' and either returns @['Category']@, or
+throws a 'CabalParseError' error. This function may also throw any one of the
+errors inherited by the use of 'downloadFile' defined in @IW.Effects.Download@.
```
vrom911: In haddock, you can actually link to the module: https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810751008
```suggestion
errors inherited by the use of 'downloadFile' defined in "IW.Effects.Download".
```

src/IW/App/Error.hs
```diff
@@ -118,7 +123,20 @@ data AppErrorType
     | UrlDownloadFailed Url
     deriving (Show, Eq)
 
--- | Map the @github@ library's @Error@ type into AppErrorType.
+type CabalErrorInfo = (Maybe Version, [CabalPError])
```
vrom911: Let's make it a data type with the record fields instead? What do you think?

src/IW/Effects/Download.hs
```diff
@@ -27,11 +27,14 @@ instance MonadDownload App where
 
 type WithDownload env m = (MonadIO m, MonadReader env m, WithError m, WithLog env m, Has Manager env)
 
+{- | This function takes an @Url@ and either returns a @ByteString@ representing
```
vrom911:
```suggestion
{- | This function takes an 'Url' and either returns a 'ByteString' representing
```

src/IW/Effects/Download.hs
```diff
@@ -27,11 +27,14 @@ instance MonadDownload App where
...
+the file contents, or throws an @UrlDownloadFailed@ error.
```
vrom911:
```suggestion
the file contents, or throws an 'UrlDownloadFailed' error.
```

src/IW/Effects/Download.hs
```diff
@@ -44,6 +47,7 @@ downloadFileImpl url@Url{..} = do
             log E $ "Couldn't download file from " <> unUrl
             throwError $ UrlDownloadFailed url
 
+-- | A verison of @downloadFile@ that returns a @Maybe BytseString@,
```
vrom911:
```suggestion
-- | A version of 'downloadFile' that returns a @'Maybe' 'BytseString'@.
```

src/IW/Effects/Cabal.hs
rashadg1030: @vrom911 Nice!

src/IW/App/Error.hs
rashadg1030: @vrom911 Looks good to me.

PR #146, opened 2019-08-23
[#112] Initial page wire up

### LINKED ISSUE #112: Wire up initial loading of issues on frontend

vrom911:
Resolves #112

The integration itself is not working at the moment, but the main idea with labels and categories (as per the first frontend drafts) is there. You can take a look and check the idea and suggest changes (not in design though, there is no design :smile: )

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
--- a/frontend/elm.json
+++ b/frontend/elm.json
@@ -6,6 +6,7 @@
     "elm-version": "0.19.0",
     "dependencies": {
         "direct": {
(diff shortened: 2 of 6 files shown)


frontend/package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
```
chshersh: Should we add this file to `.gitignore`?

chshersh (commented):
This looks great so far!

chshersh:
@rashadg1030 Try to follow the instructions to build the frontend and see how it works for you :slightly_smiling_face: After you can build frontend on your machine and make sure that it works, you can share your feedback (about instructions or something else), and we can slowly continue with the frontend further.

chshersh (approved):

PR #149, opened 2019-09-13
[#144] Update resolver to lts-14.5

### LINKED ISSUE #144: Bump up to lts-14.* resolver
This allows to cleanup `stack.yaml` significantly. Update dependencies in the `issue-wanted.cabal` file accordingly if needed.

sshine:
Resolves #144.

Updates constraints for the following packages to:

- http-client ^>= 0.6
- postgresql-simple-named ^>= 0.0
- tomland ^>= 1.1

Removes all extra-deps from stack.yaml that are present in lts-14.5.

Updates postgresql-simple-named from 0.0.0.0 to 0.0.2.0 and releases its constraint to `^>= 0.0`.

This assumes that all these packages use the PVP.

### DIFF
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -117,7 +117,7 @@ library
                      , lens ^>= 4.17
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
-                     , postgresql-simple-named ^>= 0.0
+                     , postgresql-simple-named ^>= 0.0.2
                      , random ^>= 1.1
                      , resource-pool
                      , servant ^>= 0.16
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -117,7 +117,7 @@ library
                      , lens ^>= 4.17
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
-                     , postgresql-simple-named ^>= 0.0.2
+                     , postgresql-simple-named ^>= 0.0.2.0
                      , random ^>= 1.1
                      , resource-pool
                      , servant ^>= 0.16
--- a/issue-wanted.cabal
+++ b/issue-wanted.cabal
@@ -111,20 +111,20 @@ library
                      , elm-street ^>= 0.0.1
                      , github == 0.22
                      , http-api-data ^>= 0.4
-                     , http-client ^>= 0.5.14
+                     , http-client ^>= 0.6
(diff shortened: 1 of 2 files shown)


sshine:
Let me know if any of my assumptions about the constraints I've set are wrong.

stack.yaml
```diff
@@ -1,29 +1,9 @@
-resolver: lts-13.26
+resolver: lts-14.5
```
chshersh: I need to update docker image first for this PR to be merged in `master`. Will do it soon

issue-wanted.cabal
```diff
@@ -111,20 +111,20 @@ library
...
                      , http-api-data ^>= 0.4
-                     , http-client ^>= 0.5.14
+                     , http-client ^>= 0.6
                      , http-client-tls ^>= 0.3.5.3
                      , http-types ^>= 0.12
                      , lens ^>= 4.17
                      , mtl ^>= 2.2.2
                      , postgresql-simple ^>= 0.6.1
-                     , postgresql-simple-named ^>= 0.0.0.0
+                     , postgresql-simple-named ^>= 0.0
```
vrom911: I would prefer to keep it as it is. I'm not very sure in PVP comparison operators so I prefer to restrict the lower bounds explicitly.

sshine: Do you mean as in `^>= 0.0.2`?

vrom911: To be precise, it is `^>= 0.0.2.0`.

sshine: Would that be equivalent to `== 0.0.2.0` if we assume that the package won't have more components? I noticed that some dependency constraints for packages that you own do have `==` rather than `^>=`, and I assumed it was precisely because you'd be able to and prefer to bump both, which would be necessary when fixating the constraint this much.

I'll update it to `^>= 0.0.2.0`. Always curious to learn your practices.

vrom911: It is not the same as the `==` operator. `==` means exactly that version. It differs from `^>=` because `^>= 1.2.0` desugars to `>= 1.2.0 && < 1.3` (by the second part of the PVP version component).  I prefer to specify the full version bounds because of some weird comparison rules (example `0.0` is greater than `0.0.0.0`). That's why I'm trying to avoid implicit versions where possible.

sshine: Thanks for clarifying.

